PR #1, opened 2018-02-13
Profiler eagerness

mpdeimos:


Profiler/CProfilerCallback.cpp
```diff
@@ -63,14 +61,16 @@ HRESULT CProfilerCallback::Initialize(IUnknown* pICorProfilerInfoUnkown) {
...
-		writeTupleToFile(LOG_KEY_INFO, "Mode: eager");
-	}
-	else {
-		writeTupleToFile(LOG_KEY_INFO, "Mode: lazy");
+	std::string eagernessValue = getOption("EAGERNESS");
+	if (!eagernessValue.empty()) {
+		try {
+			eagerness = std::stoi(eagernessValue);
+		} catch (std::exception e) {
+			writeTupleToFile(LOG_KEY_WARN, "Eagerness must be number that indicates the amount of method calls until traces are written: " + eagernessValue);
```
karottenreibe: add to error message: "Found instead:" before the read eagerness value

Profiler/CProfilerCallback.cpp
```diff
@@ -359,41 +347,44 @@ int CProfilerCallback::writeFileVersionInfo(LPCWSTR assemblyPath, char* buffer,
...
-		writeSingleFunctionInfoToLog(LOG_KEY_INLINED, info);
-	}
-	else {
-		// Save information about inlined method.
-		if (inlinedMethodIds.insert(calleeId).second == true) {
-			inlinedMethods.push_back(info);
+	EnterCriticalSection(&criticalSection);
+
+	if (eagerness == 1) {
+		// Directly write to log if we want to record each function immediatelly
```
karottenreibe: is this still an acutal use case that we have for any customer? otherwise i'd treat eagerness == 1 the same as any other eagerness > 0

Profiler/CProfilerCallback.cpp
```diff
@@ -359,41 +347,44 @@ int CProfilerCallback::writeFileVersionInfo(LPCWSTR assemblyPath, char* buffer,
...
+		// This is solely for performance reasons as we are writing data live
+		writeSingleFunctionInfoToLog(logKey, info);
+	} else {
+		// otherwise record function info
+		recordedFunctionInfos->push_back(info);
```
karottenreibe: these inline comments (also below) seem redundant. please remove. for the one below: how about extracting a method for the condition check with a speaking name?

Profiler/CProfilerCallback.h
```diff
@@ -84,10 +78,10 @@ class CProfilerCallback : public CProfilerCallbackBase {
 	bool isLightMode = false;
 
 	/** 
-	 * Whether to run in eager mode and write all invocations to the trace 
-	 * file directly instead of waiting until shutdown. 
+	 * Whether to run in eager mode and write a batch of recorded invocations to the trace
```
karottenreibe: please mention that the batch size is equal to the number stored here

Profiler/CProfilerCallback.cpp
mpdeimos: as mentioned, this is solely for performance reasons. removed this now. if you think we should keep, please let me know.

Profiler/CProfilerCallback.cpp
mpdeimos: done

Profiler/CProfilerCallback.cpp
mpdeimos: good idea

Profiler/CProfilerCallback.h
mpdeimos: done

mpdeimos:
@karottenreibe  RFR again. Re-introduced locking for profiling callbacks from .NET runtime.

PR #8, opened 2018-02-19
Log App Pool ID if available

### LINKED ISSUE #7: Log IIS Application Pool ID
When running a .NET web application in IIS, the process is always w3wp.exe, so we cannot distinguish between multiple applications on the same server.

The easiest way to accomplish that would be to log the environment variable APP_POOL_ID into the trace file, since this contains the application pool ID, which can be used to uniquely identify the application.

agoeb:
Closes #7 

- [x] Tests written
- [x] Documentation updated


Profiler/CProfilerCallback.cpp
```diff
@@ -36,9 +36,9 @@ namespace {
 	/** The version of the profiler */
 	const char* PROFILER_VERSION_INFO =
 #ifdef _WIN64
-		"Coverage profiler version 0.11.0 (64bit)"
+		"Coverage profiler version 0.11.1 (64bit)"
```
mpdeimos: Actually I'd like to switch to versioning with YY.MM.xx like Christian proposed, e.g. 18.2.0. WOuld also be cool to get the version injected from the build process. What do you think?

I'll have a look at injecting later...

mpdeimos: See https://github.com/cqse/teamscale-profiler-dotnet/pull/9 :)

agoeb: Done :)

agoeb:
I Don't think any of the installation instructions for IIS is obsolete with the changes I did, so I would like to keep them all. However, I added a paragraph explaining that the App Pool ID is logged in the trace file, and what that might be useful for.

PR #10, opened 2018-02-22
Process name

### LINKED ISSUE #6: Make Profiled Process Configurable
In cases where the profiled application is not executed using a dedicated user account, we might need to configure the profiler system-globally, which will also profile all other .NET processes on that machine, e.g. random powershell scripts for maintenance that may run as frequently as every minute. So we get lots of worthless traces.

The profiler should have an option to provide the name of an executable (e.g. w3wp.exe) in the config file or environment variable. Only executions of that executable should produce traces, for all others the profiler should not do anything.

Otherwise we might get tons of irrelevant traces.

agoeb:
Closes #6.

- [x] Tests written
- [x] Documentation updated


Profiler/CProfilerCallback.cpp
```diff
@@ -43,7 +43,24 @@ CProfilerCallback::~CProfilerCallback() {
 	DeleteCriticalSection(&criticalSection);
 }
 
+/** Whether the given value ends with the given suffix. */
+inline bool endsWith(std::string const & value, std::string const & suffix)
+{
+	if (suffix.size() > value.size()) return false;
```
mpdeimos: wrap in curly brackets

Profiler/CProfilerCallback.cpp
```diff
@@ -43,7 +43,24 @@ CProfilerCallback::~CProfilerCallback() {
...
+/** Whether the given value ends with the given suffix. */
+inline bool endsWith(std::string const & value, std::string const & suffix)
+{
+	if (suffix.size() > value.size()) return false;
+	return std::equal(suffix.rbegin(), suffix.rend(), value.rbegin());
+}
+
 HRESULT CProfilerCallback::Initialize(IUnknown* pICorProfilerInfoUnkown) {
+	std::string process = getProcessInfo();
+	std::string processToProfile = getEnvironmentVariable("PROCESS");
```
mpdeimos: Maybe rename getEnvironmentVariable to getConfigValue?

Profiler_Test/ProfilerTest.cs
```diff
@@ -13,7 +13,19 @@ namespace Cqse.Teamscale.Profiler.Dotnet
...
+		/// Makes sure that processes not matching the given process name are not profiled.
+		/// </summary>
+		[TestCase("w3wp.exe", 0)]
+		[TestCase("ProfilerTestee.exe", 1)]
+		[TestCase("profilerTesTEE.EXE", 1)]
+		public void TestProcessSelection(string process, int expectedTraces)
+		{
+			List<FileInfo> traces = RunProfiler("ProfilerTestee.exe", arguments: "none", lightMode: true, bitness: Bitness.x86, environment: new Dictionary<string, string> { { "COR_PROFILER_PROCESS", process } });
+			Assert.That(traces, Has.Count.EqualTo(expectedTraces), "Got the wrong number of traces.");
+		}
```
mpdeimos: You can return an int from the function and expect that with Result=1 in the annotation

Profiler/CProfilerCallback.cpp
agoeb: In order to distinguish it from the config options that can also come from the config file, we need to be a bit more specific, but rename was a good idea, since I struggled last time with the fact that it assumes the COR_PROFILER_ prefix.

Profiler_Test/ProfilerTest.cs
agoeb: Seems they changed that to ExpectedResult

agoeb:
Thanks - that was valuable feedback. Regarding the utils class: I fully agree that would make sense. But we need to think a bit about what exactly to extract, so I created a new issue  #11 for that.

PR #17, opened 2018-07-16
Remove useless code from retrieving the function info data

mpdeimos:
Retrieving the function info data contained some code that should not be needed anymore. I guess this was once needed to also display the assembly or method name in the trace. But as we use assembly ids now, most of it can be removed.

One could also think of directly using the .net provided assembly id instead of providing an own mapping?

Tested the modification with diffing the trace file of one of our tests and the result looks reasonable.

[coverage_20180716_1408310030.txt](https://github.com/cqse/teamscale-profiler-dotnet/files/2198013/coverage_20180716_1408310030.txt)
[coverage_20180716_1415480335.txt](https://github.com/cqse/teamscale-profiler-dotnet/files/2198014/coverage_20180716_1415480335.txt)

### DIFF
--- a/Profiler/CProfilerCallback.cpp
+++ b/Profiler/CProfilerCallback.cpp
@@ -270,8 +270,6 @@ HRESULT CProfilerCallback::AssemblyLoadFinished(AssemblyID assemblyId, HRESULT h
 		return S_OK;
 	}
 
-	int assemblyNumber = registerAssembly(assemblyId);
-
 	char assemblyInfo[BUFFER_SIZE];
 	int writtenChars = 0;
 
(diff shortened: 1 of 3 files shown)


mpdeimos:
Here is another test, removing the assembly id normalization (not included in the PR yet).
[coverage_20180716_1415480335.txt](https://github.com/cqse/teamscale-profiler-dotnet/files/2198258/coverage_20180716_1415480335.txt)
[coverage_20180716_1530440298.txt](https://github.com/cqse/teamscale-profiler-dotnet/files/2198259/coverage_20180716_1530440298.txt)

mpdeimos:
Integrated w/ #1

PR #19, opened 2018-09-03
Gui

### LINKED ISSUE #18: Move profiler GUI to this repo
Currently in a separate GitLab project.

We should also take care to remove the custom file copying logic from the UI since that can be handled by the uploader now. Instead the GUI should have an option to set the UPLOAD_DEAMON env variable and possibly a UI for editing the UploadDaemon.json

karottenreibe:
Fixes #18  .

- [ ] Tests written
- [ ] Documentation updated


karottenreibe:
Suggested Documentation:

---

# GUI

To configure and launch the profiler, you can use the `GUI\ProfilerGUI.exe`. It lets you select the profiled application and the target directory to which traces should be written. From there you can launch the application with the profiler.

The GUI supports parsing `.lnk` shortcut files as well, i.e. it will read the executable, command line arguments and working directory from the `.lnk` files.

The configured values can be saved to disk. If you start the GUI with the command line argument `launchTargetApp`, the saved values will be used to directly launch the target application with the profiler attached. This is useful, e.g. to provide a one-click shortcut on the Desktop for testers to start their application with the profiler.

The GUI writes a log file to `GUI\ProfilerGUI.log`. Logging can be configured with `GUI\nlog.config`.

karottenreibe:
rfr

karottenreibe:
I added code to also allow users to configure the upload daemon via the UI

karottenreibe:
reassigning to daniel

ProfilerGUI/Source/Configurator/BoolToCollapsingConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Windows;
+using System.Windows.Data;
+
+namespace ProfilerGUI.Source.Configurator
+{
+    /// <summary>
+    /// Converts a boolean to either Visibility.Visible or Visibility.Collapsed
+    /// </summary>
+    [ValueConversion(typeof(bool), typeof(Visibility))]
+    public class BoolToCollapsingConverter : IValueConverter
```
mpdeimos: Use .net internal class BooleanToVisibilityConverter

ProfilerGUI/Source/Configurator/MachineTypeExtractor.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// </summary>
+    internal class MachineTypeUtils
+    {
+        private const int ExpectedPEHead = 0x00004550;
+
+        /// <summary>
+        /// Tries to find the "bitness" (32bit vs 64bit) of the given executable by looking at the PE header.
+        /// This is needed to select the correct profiler DLL to run the application against.
+        ///
+        /// Taken from https://stackoverflow.com/questions/1001404/check-if-unmanaged-dll-is-32-bit-or-64-bit
```
mpdeimos: Isn't the article regarding unmanaged code?

ProfilerGUI/Source/Configurator/MainViewModel.cs
```diff
@@ -0,0 +1,190 @@
...
+        private UploadConfig ReadUploadConfigFromDisk()
+        {
+            if (!File.Exists(UploadConfigFilePath))
+            {
+                return null;
+            }
+
+            try
+            {
+                return JsonConvert.DeserializeObject<UploadConfig>(File.ReadAllText(UploadConfigFilePath));
```
mpdeimos: Add UploadConfig.FromFile()?

ProfilerGUI/Source/Configurator/MainViewModel.cs
```diff
@@ -0,0 +1,190 @@
...
+namespace ProfilerGUI.Source.Configurator
+{
+    /// <summary>
+    /// The main view model to interact with the <see cref="MainWindow"/>.
+    /// </summary>
+    public class MainViewModel : INotifyPropertyChanged
+    {
+        private static readonly Logger logger = LogManager.GetCurrentClassLogger();
+
+        private static readonly string UploadConfigFilePath = Path.Combine(Directory.GetParent(AppDomain.CurrentDomain.BaseDirectory).FullName, "UploadDaemon", UploadConfig.ConfigFileName);
```
mpdeimos: we're writing to this file, hence I would not using the install dir but something in APP_DATA. If we want support distributing config files w/ profiles we should allow opening them.

ProfilerGUI/Source/Configurator/UploadViewModel.cs
```diff
@@ -0,0 +1,193 @@
...
+            {
+                Config.Directory = value;
+                PropertyChanged.Raise(this);
+            }
+        }
+
+        /// <summary>
+        /// Whether the configuration UI for the Teamscale server should be shown.
+        /// </summary>
+        public bool IsTeamscaleConfigVisible { get => Config != null && Config.Teamscale != null; }
```
mpdeimos: IsTeamscaleConfigVisible => Config != ...
no need to use curlies and get :)

ProfilerGUI/Source/Configurator/UploadViewModel.cs
```diff
@@ -0,0 +1,193 @@
...
+            }
+
+            return await CheckTeamscaleConnection();
+        }
+
+        /// <summary>
+        /// Checks if the connection to Teamscale can be established.
+        /// Shows an error message in the UI if not.
+        /// </summary>
+        private async Task<bool> CheckTeamscaleConnection()
```
mpdeimos: We're now duplicating functionality from the teamscale client lib. I think it's a good time to integrate it now and add features there if needed.

ProfilerGUI/Source/Configurator/WindowsShortcutReader.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            WshShell shell = new WshShell();
+            IWshShortcut link = (IWshShortcut)shell.CreateShortcut(filePath);
+            return new ShortcutInfo(link.TargetPath, link.WorkingDirectory, link.Arguments);
+        }
+
+        /// <summary>
+        /// Result of parsing a .lnk file.
+        /// </summary>
+        public class ShortcutInfo
```
mpdeimos: ditch helper class shortcutreader and add constructor/factory method ShortcutInfo.Read()

mpdeimos: ...and ShortcutInfo is a really bad name. why not simply Shurtcut?

ProfilerGUI/Source/Configurator/WpfUtils.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// Opens a folder chooser and calls the given action with the selected folder if the user chose one.
+        /// </summary>
+        public static void OpenFolderChooser(Action<string> pathSelectedAction)
+        {
+            using (FolderBrowserDialog dialog = new FolderBrowserDialog())
+            {
+                DialogResult result = dialog.ShowDialog();
+                if (result == System.Windows.Forms.DialogResult.OK)
+                {
+                    pathSelectedAction.Invoke(dialog.SelectedPath);
```
mpdeimos: I think you can remove the .Invoke part

ProfilerGUI/Source/Configurator/WpfUtils.cs
```diff
@@ -0,0 +1,55 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Raises the OnPropertyChangedEvent of an INotifyPropertyChanged object.
+        /// </summary>
+        public static void Raise(this PropertyChangedEventHandler handler, INotifyPropertyChanged sender, [CallerMemberName] string propertyName = "")
```
mpdeimos: This method is not Wpf-Specific, hence needs to be placed somewhere else.

mpdeimos: afaik, our guidelines do not allow extension methods.

ProfilerGUI/Source/Shared/ProcessOutput.cs
```diff
@@ -0,0 +1,35 @@
...
+        public string StdErr { get; }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        public ProcessResult(int returnCode, string stdout = null, string stderr = null)
+        {
+            ReturnCode = returnCode;
+            StdOut = stdout ?? string.Empty;
+            StdErr = stderr ?? string.Empty;
```
mpdeimos: use empty string already as default param?

ProfilerGUI/Source/Shared/ProfilerConfiguration.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        /// <summary>
+        /// The name of the config file.
+        /// </summary>
+        private const string ConfigFileName = "ProfilerGUI.json";
+
+        /// <summary>
+        /// Path to the config file.
+        /// </summary>
+        public static readonly string ConfigFilePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ConfigFileName);
```
mpdeimos: Again, we write to this file, hence I'd place in APP_DATA. I'm also pretty sure that we might need more than one config saveable.

mpdeimos: Actually it would be nice to have one config file for uploader and daemon

mpdeimos: Any reason we we're not re-using the already existing config file for the profiler?

ProfilerGUI/Source/Shared/ProfilerConstants.cs
```diff
@@ -0,0 +1,48 @@
+﻿namespace ProfilerGUI.Source.Shared
+{
+    /// <summary>
+    /// Constants used when running the profiler.
+    /// </summary>
+    public static class ProfilerConstants
```
mpdeimos: This class does not handle .NET Core variable names

ProfilerGUI/Source/Shared/SystemUtils.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            ProcessStartInfo startInfo = new ProcessStartInfo
+            {
+                FileName = processPath,
+                Arguments = args
+            };
+
+            startInfo.UseShellExecute = false;
+
+            foreach (var (environmentVariable, value) in environmentVariables)
```
mpdeimos: i'd explicitly write (string .., string..) according to guidelines

ProfilerGUI/Source/Shared/SystemUtils.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                startInfo.EnvironmentVariables.Remove(environmentVariable);
```
mpdeimos: document why remove is called

ProfilerGUI/UploadConfigWindow.xaml.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// changes to the upload config or not.
+    /// </summary>
+    public partial class UploadConfigWindow : Window
+    {
+        private readonly UploadViewModel ViewModel;
+
+        /// <summary>
+        /// The config the user edited.
+        /// </summary>
+        public UploadConfig Config { get => ViewModel.Config; }
```
mpdeimos: use shorthand Config => ...

ProfilerGUI/UploadConfigWindow.xaml.cs
```diff
@@ -0,0 +1,63 @@
...
+        public UploadConfigWindow(UploadConfig config)
+        {
+            InitializeComponent();
+            ViewModel = new UploadViewModel(config?.Clone());
+            this.DataContext = ViewModel;
+        }
+
+        private void OnValidateClicked(object sender, RoutedEventArgs e)
+        {
+            Task<bool> _ = ViewModel.Validate();
```
mpdeimos: remove type arg

Profiler_Test/ProfilerTest.cs
```diff
@@ -8,82 +8,81 @@
 
 namespace Cqse.Teamscale.Profiler.Dotnet
 {
```
mpdeimos: Please undo formatting changes, as this makes merging hell. Afaik we should use tabs anyways...

ProfilerGUI/Helper/WpfRichTextBoxRowColoringRule.cs
```diff
@@ -0,0 +1,97 @@
...
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without 
+// modification, are permitted provided that the following conditions 
+// are met:
+// 
+// * Redistributions of source code must retain the above copyright notice, 
+//   this list of conditions and the following disclaimer. 
+// 
+// * Redistributions in binary form must reproduce the above copyright notice,
```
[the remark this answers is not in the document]
karottenreibe: it's a bit weird, installing the nuget package puts this code in the solution (?) I wasn't sure if this is standard practice in .NET

ProfilerGUI/MainWindow.xaml.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            WpfUtils.OpenFolderChooser(selectedPath => ViewModel.TargetApp.TraceDirectory = selectedPath);
+        }
+
+        private void OnChooseApplicationButtonClick(object sender, EventArgs args)
+        {
+            using (OpenFileDialog dialog = new OpenFileDialog
+            {
+                // This lets the user select lnk-files (shortcuts), which can be helpful in our context
+                DereferenceLinks = false,
```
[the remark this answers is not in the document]
karottenreibe: no, because then we get the file the link points to, not the link itself

ProfilerGUI/Source/Configurator/BoolToCollapsingConverter.cs
karottenreibe: as far as I understood, that converts to invisible, not to collapsed. Can you show me how to use the builtin one to get the collapsed state?

ProfilerGUI/Source/Configurator/MachineTypeExtractor.cs
karottenreibe: I tested this against a normal .NET executable and it selected the correct bitness

Cqse.Teamscale.Profiler.Dotnet.sln
```diff
@@ -7,30 +7,90 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Profiler", "Profiler\Profil
...
 Global
+	GlobalSection(SharedMSBuildProjectFiles) = preSolution
+		Common\Common.projitems*{0bb237cf-2c22-48dc-95e3-e9f58ab6a398}*SharedItemsImports = 4
+		Common\Common.projitems*{a01a7519-e10d-4153-8cd4-678935721df3}*SharedItemsImports = 4
+		Common\Common.projitems*{ac9b3cca-44b4-429a-bec8-c516a6013cc7}*SharedItemsImports = 13
+	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
 		Debug|Win32 = Debug|Win32
 		Debug|x64 = Debug|x64
```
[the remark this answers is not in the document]
karottenreibe: done

ProfilerGUI/Helper/README_NLog_WpfRichTextBox.txt
```diff
@@ -0,0 +1,28 @@
+﻿EXAMPLES AND USAGE
+
+NOTE: Currently must be configured via code
+
```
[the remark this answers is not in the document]
karottenreibe: done

ProfilerGUI/Source/Configurator/BoolToCollapsingConverter.cs
karottenreibe: ok looks like I misunderstood ;-) removed this

ProfilerGUI/Source/Configurator/MainViewModel.cs
karottenreibe: done

ProfilerGUI/Source/Configurator/MainViewModel.cs
karottenreibe: created #29 for this. I'm not 100% sure this is what we want. Let's discuss this after merging this

ProfilerGUI/Source/Configurator/UploadViewModel.cs
karottenreibe: didn't know that! :-) done

ProfilerGUI/Source/Configurator/UploadViewModel.cs
karottenreibe: created #30 for this. As far as I remember from our last discussion, we'd first need to turn that into a nuget package and then still have to rewrite this code here, push some stuff to the lib and retest the entire thing. I'd rather get this merged soon so I can use this code in the next pilot and do the maintenance changes afterwards without the time pressure

ProfilerGUI/Source/Configurator/WindowsShortcutReader.cs
karottenreibe: done

ProfilerGUI/Source/Configurator/WpfUtils.cs
karottenreibe: done

ProfilerGUI/Source/Configurator/WpfUtils.cs
karottenreibe: renamed the file to be more generic

karottenreibe: done

ProfilerGUI/Source/Shared/ProcessOutput.cs
karottenreibe: I had to use a literal `""` then. I wasn't sure if that was the canonical way to do it at the time I wrote the code

ProfilerGUI/Source/Shared/ProfilerConfiguration.cs
karottenreibe: the existing config file is very simplistic on purpose (properties-like) to be able to parse it easily in C++. The uploader is a separate process and I didn't want to shoehorn the much more complicated config into this simple format. And also I didn't want to parse JSON in C++ for obvious reasons :P 

As to APP_DATA, see #29

ProfilerGUI/Source/Shared/ProfilerConstants.cs
karottenreibe: created #31 for that

ProfilerGUI/Source/Shared/SystemUtils.cs
karottenreibe: thanks.I tried that but all my attempts were refused by the compiler so I gave up and used var ;-)

ProfilerGUI/Source/Shared/SystemUtils.cs
karottenreibe: .NET APIs being .NET APIs... ;-)

ProfilerGUI/UploadConfigWindow.xaml.cs
karottenreibe: done

ProfilerGUI/UploadConfigWindow.xaml.cs
karottenreibe: didn't know that was possible!

Profiler_Test/ProfilerTest.cs
karottenreibe: reverted

karottenreibe:
rfr again

ProfilerGUI/Source/Runner/ProfilerRunner.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        private string LocateProfilerDll()
+        {
+            string profilerDll = ProfilerConstants.ProfilerDll64Bit;
+            if (configuration.ApplicationType == EApplicationType.Type32Bit)
+            {
+                profilerDll = ProfilerConstants.ProfilerDll32Bit;
+            }
+            return Directory.GetParent(Environment.CurrentDirectory) + @"\" + profilerDll;
```
salsolatragus: Is it intentional that this is the parent of the current directory? It somewhat surprised me that it was looking for the DLL there. I would've expected it to either look in the working directory or to let me specify a path to the DLLs (possible with a default path). I may also just be blissfully unaware as to why this makes sense in how this profiler is usually used 😇

ProfilerGUI/Source/Configurator/MachineTypeExtractor.cs
salsolatragus: I just tried the GUI with `test-data/test-programs/GeneratedTest.exe`, which is detected as 32bit. When I execute with this setting no trace is generated. When I manually switch to 64bit and run I get a trace file. I guess the detection fails, because `GeneratedTest.exe` is managed .NET code and bitness depends on the runtime environment, i.e., cannot be determined from the exe alone?

ProfilerGUI/Source/Runner/ProfilerRunner.cs
mpdeimos: I think that's right. The profiler is deployed together with the Gui with the latter being in a subdirectory. So should be OK :angel:

salsolatragus: I see. Requires manual copying of the profiler dlls when running from Visual Studio for debugging, but I guess that's acceptable in a dev environment.

ProfilerGUI/Source/Configurator/MachineTypeExtractor.cs
karottenreibe: Fair point, though in the case of a managed assembly we can still
- check if it has a fixed bitness (this is possible)
- if not, use the bitness of the machine (this is probably the correct bitness then, unless the user uses "prefer 32bit" in their project config explicitly, but then: whatever, let them deal with it ;-))

karottenreibe:
rfr. The tests run locally on my machine. @mpdeimos  do you have any idea why they would fail on the build server? It seems like no traces are being generated for some reason

ProfilerGUI/Source/Configurator/MachineTypeExtractor.cs
salsolatragus: Sounds fine to me :)

salsolatragus:
@karottenreibe I have that problem (no trace being generated) with all profiler tests locally on my machine as well. Attributed this to a problem with my local setup, but maybe it isn't? Anyways, I can reproduce this, which might help debugging...

karottenreibe:
@salsolatragus awesome! could you please run the tests for me and then check your event log (section Windows > Applications) for errors? best clear that log before the test run to make sure you know which events are related to the test. Maybe these tell us something! Thanks!

karottenreibe:
fixed the tests. Cause was an incorrectly manually edited sln file that assigned the Debug build to the Release configuration.

We shouldn't be editing around in these files manually anymore IMO

rfr again

ProfilerGUI/Source/Configurator/MainViewModel.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// <summary>
+        /// Saves the current configuration to a file.
+        /// </summary>
+        internal void SaveConfiguration()
+        {
+            try
+            {
+                TargetApp.Configuration.WriteToFile();
+                logger.Info("Wrote config to {configPath}", ProfilerConfiguration.ConfigFilePath);
+                SaveUploadConfig();
```
salsolatragus: At first I though the error message below is incomplete, because it doesn't cover the "failed to save upload config" case. Then I realised that `SaveUploadConfig()` does it's own error handling. Move the call out of the try catch?

Also: Inconsistent naming `Config` vs. `Configuration`

ProfilerGUI/Source/Configurator/MainViewModel.cs
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        /// <summary>
+        /// Starts the profiled application with the profiler attached.
+        /// </summary>
+        internal void RunProfiledApplication()
+        {
+            if (string.IsNullOrEmpty(TargetApp.ApplicationPath))
+            {
+                logger.Error("You did not configure an application to profile");
```
salsolatragus: Be instructive? "Please configure/select an application..."

Naming inconsistency: "application" vs. "Select app" on the GUI button.

ProfilerGUI/Source/Configurator/UploadViewModel.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        /// <summary>
+        /// Whether the configuration UI for the Teamscale server should be shown.
+        ///
+        /// The three cases are modeled as follows:
+        /// - no upload: Config == null
+        /// - upload to Teamscale: Config != null && Config.Teamscale != null
+        /// - upload to directory: Config != null && Config.Teamcsale == null
+        /// </summary>
+        public int UploadTypeIndex
```
salsolatragus: Azure is missing

ProfilerGUI/Source/Configurator/UploadViewModel.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        /// <summary>
+        /// Whether the configuration UI for the version assembly should be shown.
+        /// </summary>
+        public bool IsVersionAssemblyVisible => Config != null;
+
+        private ValidationResult internalValidationResult = null;
+
+        /// <summary>
+        /// The general error message.
```
salsolatragus: This documentation seems inconsistent with the property.

ProfilerGUI/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,55 @@
...
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("ProfilerGUI")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("ProfilerGUI")]
+[assembly: AssemblyCopyright("Copyright ©  2018")]
```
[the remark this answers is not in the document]
karottenreibe: unfortunately it can't be put in a central location except with some ugly VS tricks which I don't like: https://stackoverflow.com/questions/6771694/shared-assemblyinfo-for-uniform-versioning-across-the-solution

ProfilerGUI/Source/Configurator/MainViewModel.cs
```diff
@@ -0,0 +1,190 @@
...
+            {
+                if (uploadConfig == null)
+                {
+                    return "No upload";
+                }
+                if (uploadConfig.Teamscale != null)
+                {
+                    return $"Upload to {uploadConfig.Teamscale}";
+                }
+                return $"Upload to directory {uploadConfig.Directory}";
```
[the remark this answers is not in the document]
karottenreibe: you didn't merge the azure stuff into this branch AFAICS so it's not supposed to be here ;-) you only merged the .net core stuff

Common/UploadConfig.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// <summary>
+        /// The assembly from which to read the version number.
+        /// </summary>
+        public string VersionAssembly { get; set; }
+
+        /// <summary>
+        /// The directory to upload the traces to.
+        /// </summary>
+        public string Directory { get; set; } = null;
+
```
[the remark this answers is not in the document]
karottenreibe: see above

ProfilerGUI/Source/Configurator/UploadViewModel.cs
karottenreibe: see above

karottenreibe:
addressed your comments so far :-)

ProfilerGUI/ProfilerGUI.exe.manifest
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
+  <assemblyIdentity version="1.0.0.0"
+     processorArchitecture="X86"
+     name="ProfilerGUI"
```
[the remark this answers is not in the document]
salsolatragus: You're right. My bad.

ProfilerGUI/Properties/AssemblyInfo.cs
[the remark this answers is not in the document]
salsolatragus: For GUI/UploadDaemon: Why can't we go with one `AssemblyInfo.cs` file thats linked to both projects? That's what we do in the Teamscale VS plugin and it seems to work just fine there.

To make it work for the Profiler C++ project as well, I'd imagine we could place it in some properties file and read that when setting up the environment. I'm not sure that's technically feasible though...

ProfilerGUI/Source/Configurator/MainViewModel.cs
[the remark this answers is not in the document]
salsolatragus: The Azure stuff is on master, which you just merged with ;)

I suggest we create a new ticket for supporting Azure in the GUI, to get this PR closed as soon as possible.

Common/UploadConfig.cs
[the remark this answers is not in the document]
salsolatragus: I thought you moved the `Config` for the `UploadDaemon` to here (that's why I expected to see the Azure options, which, in fact, are in `Config`), but currently we have both, which seem very redundant. Is there a particular reason for this?

salsolatragus:
Other issues/thoughts:

* When I have a syntactically invalid `UploadDaemon.json`, the GUI displays "no config". When I then save the configuration via the GUI, my JSON file is overwritten!
* Comments/formatting in `UploadDaemon.json` are killed by load and save in GUI. Even if I didn't change anything.
* Upload dialog » "To a Directory": The directory selection does not show the current path from the config.
* Why is "Working directory" not a file chooser?
* Ask for save on quit GUI?
* When I profile an application that does not start (e.g., crash due to missing dependency, happens with `test-programs\DotNetCoreApp.dll`) I do not get a trace file, but the GUI looks exactly like after a successful run. Can we somehow distinguish such cases, to make clear it's not our fault there is not trace?

ProfilerGUI/Source/Configurator/MainViewModel.cs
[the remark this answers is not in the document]
karottenreibe: agree. please create the ticket if you feel we should integrate this in the GUI. We should, however, have a clear use-case for this IMO. At the moment it seems to me that it's unrealistic that anyone would want to combine these two features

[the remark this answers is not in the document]
salsolatragus: Fair enough.

What bothers me is the case where someone has an existing Azure or HttpUpload configuration and opens the GUI. In these cases, the GUI will assume there is a FileSystem configuration (`uploadConfig.Directory` will be `null`) and not display the actual configuration. I think we should at least display that there is a configuration type that is unsupported by the GUI. Created #68 for this.

Common/UploadConfig.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Common
+{
+    /// <summary>
+    /// Data class that is deserialized from the JSON configuration file.
+    /// </summary>
+    public class UploadConfig
```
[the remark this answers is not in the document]
karottenreibe: sorry for the botched merge. fixed now

karottenreibe:
rfr again. Thanks for the many good points. I'm aggressively moving further "features" for the GUI to new tickets to get this merged as you suggest - including some usability bugs.

Regarding your misc points:

> When I have a syntactically invalid UploadDaemon.json, the GUI displays "no config". When I then save the configuration via the GUI, my JSON file is overwritten!

I'd fix the GUI display but saving the config works as expected IMO. After all: there's nothing we can do when the user feeds us invalid JSON, so overwriting on save seems legit to me. Please create an issue for the GUI display

>    Comments/formatting in UploadDaemon.json are killed by load and save in GUI. Even if I didn't change anything.

Should not be killed by load. Please create a ticket for that. Killing on save is fine by me since trying to preserve comments in there programmatically is way too much overhead. IMO you either edit manually or use the GUI, never both. I don't think we have to support both at the same time.

>    Upload dialog » "To a Directory": The directory selection does not show the current path from the config.

Please create a bug for this

>    Why is "Working directory" not a file chooser?

Please create an issue for this

>    Ask for save on quit GUI?

Please create an issue for this

>    When I profile an application that does not start (e.g., crash due to missing dependency, happens with test-programs\DotNetCoreApp.dll) I do not get a trace file, but the GUI looks exactly like after a successful run. Can we somehow distinguish such cases, to make clear it's not our fault there is not trace?

Unfortunately, I don't  think we can distinguish crash + no trace from crash + trace. Unless you have a concrete idea I'd leave it as is.

salsolatragus:
I've created tasks for all the open issues and inserted your suggested documentation into the user guide. Thanks for the great work!

PR #46, opened 2018-10-12
Gui net core

### LINKED ISSUE #31: Profiler GUI should be able to handle .NET Core applications
we need a checkbox to switch between core and framework in the UI and use different constant names in that case. see the documentation about more info about the constants

salsolatragus:
Fixes #31.

- [x] Tests written
- [ ] Documentation updated

I didn't write tests for how the environment variables are constructed in the `ProfilerRunner`, because there weren't tests for the previous construction and I'm honestly not sure how to best test this environment.


ProfilerGUI/Source/Configurator/TargetAppModel.cs
```diff
@@ -120,21 +121,51 @@ private void UpdateExecutableMachineType()
...
-            }
-
-            if (typeOfExecutable == MachineTypeUtils.MachineType.I386)
-            {
-                ApplicationType = EApplicationType.Type32Bit;
-            }
-            else
-            {
-                ApplicationType = EApplicationType.Type64Bit;
+            if (IsDotNETCoreApplication(ApplicationPath))
```
salsolatragus: I implemented this separately from `MachineTypeUtils.GetExecutableType`, because .NET Core isn't really a machine type. We could also merge these two heuristics to directly determine the application type. I have no clear preference for either option.

ProfilerGUI/Source/Runner/ProfilerRunner.cs
```diff
@@ -59,16 +55,54 @@ public void RunAsynchronously()
...
+        private static string GetProfilerDllPath(string profilerDll)
+        {
+            var path = Directory.GetParent(Environment.CurrentDirectory) + @"\" + profilerDll;
+            if (!File.Exists(path))
             {
-                profilerDll = ProfilerConstants.ProfilerDll32Bit;
-            }
-            return Directory.GetParent(Environment.CurrentDirectory) + @"\" + profilerDll;
-        }
+                throw new ArgumentException(string.Format("The profiler DLL was not found at \"{0}\". Cannot profile the application", path));
```
salsolatragus: I'm not 100% happy with this string formatting, but exceptions can't do it for us like the logger can and I wanted to have the explicit `logger.Error` in `RunAsynchronously()`, to make it explicit there that we report the error before aborting the run.

ProfilerGUI/Source/Configurator/TargetAppModel.cs
mpdeimos: Sounds legit :+1:

ProfilerGUI/Source/Runner/ProfilerRunner.cs
mpdeimos: You could do string interpolation `@"foo {path} bar"`

ProfilerGUI/Source/Configurator/EApplicationType.cs
```diff
@@ -1,10 +1,10 @@
 ﻿namespace ProfilerGUI.Source.Configurator
 {
     /// <summary>
-    /// The two different modes in which the profiler can be launched: 32bit vs 64bit.
+    /// The three different modes in which the profiler can be launched: 32bit, 64bit, and .NET Core.
```
mpdeimos: do not enumerate her at all?

ProfilerGUI/Source/Configurator/TargetAppModel.cs
```diff
@@ -120,21 +121,51 @@ private void UpdateExecutableMachineType()
...
+                }
             }
+        }
+        
+        private static bool IsDotNETCoreApplication(string applicationPath)
+        {
+            try
+            {
+                // If the application is .NET, we can load it as an assembly.
+                var assembly = Assembly.LoadFile(applicationPath);
```
mpdeimos: Maybe it would be better to use Mono.Cecil (http://cecil.pe/) instead of `Assembly.Load` here? Afaik the latter will also load dependent assemblies in RAM / current VM

ProfilerGUI/Source/Runner/ProfilerRunner.cs
```diff
@@ -31,21 +31,17 @@ public ProfilerRunner(ProfilerConfiguration configuration)
...
-                    (ProfilerConstants.TargetDirectoryEnvironmentVariable, configuration.TraceTargetFolder),
-                    (ProfilerConstants.LightModeEnvironmentVariable, "1"),
-                };
+            try
+            {
+                environmentVariables = DetermineEnvironmentVariables(configuration);
+            }
+            catch (Exception e)
+            {
+                logger.Error(e.Message);
```
mpdeimos: maybe add some message where the error occurred (loading env vars). On the other hand I am not sure if we could sift down in below try..catch?

ProfilerGUI/Source/Runner/ProfilerRunner.cs
```diff
@@ -59,16 +55,54 @@ public void RunAsynchronously()
...
-            string profilerDll = ProfilerConstants.ProfilerDll64Bit;
-            if (configuration.ApplicationType == EApplicationType.Type32Bit)
+        }
+
+        private static List<(string, string)> DetermineEnvironmentVariables(ProfilerConfiguration configuration)
+        {
+            List<(string, string)> environmentVariables;
+            if (configuration.ApplicationType == EApplicationType.TypeCORE)
+            {
+                environmentVariables = new List<(string, string)>()
```
mpdeimos: Usually we prefer immediate `return`. You can also initialize the variable dictionary with the variables you add as the last step

ProfilerGUI/Source/Configurator/TargetAppModel.cs
salsolatragus: I switched to using [`Assembly.ReflectionOnlyLoadFrom`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.reflectiononlyloadfrom?view=netframework-4.7.2), since it saves us the additional dependency, avoids that dependencies are loaded, and prevents execution of any code from the assembly.

ProfilerGUI/Source/Runner/ProfilerRunner.cs
salsolatragus: I merged the two tries, looks much cleaner and makes perfect sense.

ProfilerGUI/Source/Runner/ProfilerRunner.cs
salsolatragus: I split the collection of environment variables into two functions, because they do conceptually different things (hooking up the profiler in the environment vs. configuring the profiler itself). Does this work for you?

salsolatragus:
rtr

When I installed the Teamscale vssettings and activated format on save, I got tons of formatting changes. I didn't commit these yet, because they completely obfuscate the changes I made to address the review. How do we best deal with this?

ProfilerGUI/Source/Configurator/TargetAppModel.cs
```diff
@@ -149,8 +149,10 @@ private static bool IsDotNETCoreApplication(string applicationPath)
         {
             try
             {
-                // If the application is .NET, we can load it as an assembly.
-                var assembly = Assembly.LoadFile(applicationPath);
+				// If the application is .NET, we can load it as an assembly.
+				// We load into the reflection context here, to avoid loading dependencies
+				// and prevent any code from executing.
+				var assembly = Assembly.ReflectionOnlyLoadFrom(applicationPath);
```
mpdeimos: Good catch, wasn't aware of this method at all.

ProfilerGUI/Source/Runner/ProfilerRunner.cs
mpdeimos: Sure. reads much better now.

ProfilerGUI/Source/Configurator/TargetAppModel.cs
mpdeimos: Good catch, wasn't aware of this method at all. That'S the point where reviews make perfectly sense :)

mpdeimos:
Lookig good! Regarding formatting let's decide on the best tool (VS Internal, EditorConfig, CodeMaid) and the create a ticket for configuring & reformatting.

PR #49, opened 2018-10-25
Upload traces to an Azure File Storage

salsolatragus:
Fixes VSTS #281194.

- [x] Tests written
- [x] Documentation updated

I've considered writing tests for the upload itself, but since this is essentially a thin layer over the .NET Azure library and we would have to mock the entire library (ourselves) to write tests, I don't think this is worth it.

### DIFF
--- a/UploadDaemon/AzureFileStorage.cs
+++ b/UploadDaemon/AzureFileStorage.cs
@@ -1,0 +1,26 @@
+﻿using Newtonsoft.Json;
+
+/// <summary>
+/// Data class that holds all details necessary to connect to an Azure File Storage.
+/// </summary>
+public class AzureFileStorage
+{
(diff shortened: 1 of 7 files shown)


mpdeimos:
Cool. Looks much simpler than the Java case :+1: 

I've added some TODOs in the code (see last commit).

salsolatragus:
rfr

* The `KeyVault` dependency is required (transitively) by `Storage.Common` and `Storage.File`.
* I've implemented recursive creation of the directory, however, I'm uncertain about the use of `async/await` in the presence of recursion... From what I gathered, it shouldn't become a problem in the rather bounded scenario of nested directories, but I'm not familiar with how this works under the hood. Happy to learn more :)

PR #54, opened 2018-10-30
Stacktrace

### LINKED ISSUE #55: Write stack traces in case of profiler crashes
We recently had the profiler crash mysteriously. This is very hard to debug. We need a way to get a proper stack trace in case our code crashes e.g. with an access violation.

karottenreibe:
Fixes #55

- [ ] Tests written
- [x] Documentation updated

Since this error handling is inherently untestable, I instead tested this manually by introducing a simple segfault into the profiler. I tested both with and without the IGNORE_EXCEPTIONS variable.

I verified that the resulting stack traces are readable and contain the function names. This, however, requires that the .pdb files are present next to the .dll files

### DIFF
--- a/Profiler/CProfilerCallback.cpp
+++ b/Profiler/CProfilerCallback.cpp
@@ -12,14 +12,40 @@
 #pragma intrinsic(strcmp,labs,strcpy,_rotl,memcmp,strlen,_rotr,memcpy,_lrotl,_strset,memset,_lrotr,abs,strcat)
 
 CProfilerCallback::CProfilerCallback() {
-	InitializeCriticalSection(&callbackSynchronization);
+	try {
+		InitializeCriticalSection(&callbackSynchronization);
+	}
+	catch (...) {
+		if (!handleException("Constructor")) {
+			throw;
+		}
+	}
 }
 
 CProfilerCallback::~CProfilerCallback() {
-	DeleteCriticalSection(&callbackSynchronization);
+	try {
+		DeleteCriticalSection(&callbackSynchronization);
+	}
+	catch (...) {
+		if (!handleException("Destructor")) {
+			throw;
+		}
+	}
 }
 
 HRESULT CProfilerCallback::Initialize(IUnknown* pICorProfilerInfoUnkown) {
+	try {
+		return InitializeImplementation(pICorProfilerInfoUnkown);
+	}
+	catch (...) {
+		if (handleException("Initialize")) {
+			return S_OK;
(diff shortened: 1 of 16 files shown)


Profiler/CProfilerCallback.h
```diff
@@ -160,4 +159,13 @@ class CProfilerCallback : public CProfilerCallbackBase {
...
 	int writeFileVersionInfo(LPCWSTR moduleFileName, char* buffer, size_t bufferSize);
+
+	HRESULT JITCompilationFinishedImplementation(FunctionID functionID, HRESULT hrStatus, BOOL fIsSafeToBlock);
+	HRESULT ShutdownImplementation();
+	HRESULT AssemblyLoadFinishedImplementation(AssemblyID assemblyID, HRESULT hrStatus);
+	HRESULT JITInliningImplementation(FunctionID callerID, FunctionID calleeID, BOOL *pfShouldInline);
+	HRESULT InitializeImplementation(IUnknown *pICorProfilerInfoUnk);
+
+	/** Logs a stack trace. If true is returned, the calling code should continue, otherwise it should rethrow the exception. */
+	bool handleException(std::string context);
```
zipperle: Why not rethrow the exception in this function?

karottenreibe: unfortunately, that's not possible. I have to catch (...) in the outer function so I can catch Windows structured exceptions as well (e.g. division by 0, general protection fault, ...) since these are not represented by actual exception objects. Thus, I can only rethrow via a throw; statement. And this only works in the context of a catch statement, which I don't have here anymore :-(

karottenreibe:
rfr again @zipperle

karottenreibe:
@zipperle I fixed the rethrow as we discussed :-) 
rfr again

Profiler/CProfilerCallback.cpp
```diff
@@ -282,6 +326,24 @@ void CProfilerCallback::getAssemblyInfo(AssemblyID assemblyId, WCHAR *assemblyNa
...
+	}
+	catch (...) {
+		handleException("JITCompilationFinished");
+		return S_OK;
+	}
+}
+
+void CProfilerCallback::handleException(std::string context) {
+	Debug::logStacktrace(context);
+	if (getOption("IGNORE_EXCEPTIONS") != "1") {
```
zipperle: Can you add this parameter to the documentation?

karottenreibe:
rfr

PR #57, opened 2018-11-02
don't profile the upload daemon process

### LINKED ISSUE #56: Don't profile the upload daemon
I'm not sure but I think we currently pass on the COR_PROFILER_* variables from the profiled process to the upload deamon. This would lead to the daemon being profile, which makes little sense. Instead we should explicitly disable profiling for the daemon process.

karottenreibe:
Fixes #56 

- [ ] Tests written
- [ ] Documentation updated


karottenreibe:
rfr, tested manually on my machine, no need to update documentation

PR #64, opened 2018-11-28
auto-reload logging config and log exceptions with stack trace

### LINKED ISSUE #44: Default logging config does not log exceptions
Should be replaced with a config that does. We should also turn on auto reloading of the configs. Example config:

```xml
﻿<?xml version="1.0" encoding="utf-8" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    autoReload="true">

  <targets>
    <target name="logfile" xsi:type="File" fileName="UploadDaemon.log" maxArchiveFiles="10"
        archiveNumbering="Sequence" archiveFileName="UploadDaemon.{#######}.log" archiveAboveSize="10485760"
        layout="${longdate}|${level:uppercase=true}|${logger}|${message}|${exception:format=ShortType,Message,Stacktrace:innerFormat=ShortType,Message,Stacktrace:maxInnerExceptionLevel=10}" />
  </targets>

  <rules>
    <logger name="*" minlevel="Warn" writeTo="logfile" />
  </rules>
</nlog>
```

c.f. https://github.com/nlog/nlog/wiki/Layouts

karottenreibe:
Fixes #44 

- [ ] Tests written (not needed - tested by using the new config)
- [ ] Documentation updated (not needed)

### DIFF
--- a/UploadDaemon/nlog.config
+++ b/UploadDaemon/nlog.config
@@ -1,9 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
-      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      autoReload="true">
 
   <targets>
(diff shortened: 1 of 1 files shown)


karottenreibe:
rfr

PR #77, opened 2018-12-10
Build output

### LINKED ISSUE #76: Various improvements to local development workflow
- building locally should output files in the same dir structure as the dist to allow running the GUI and uploader directly without copying files
- we should treat warnings as errors
- it should be possible to build both profiler DLLs in one go from VS

karottenreibe:
Fixes #76 

- [ ] Tests written (not needed)
- [x] Documentation updated


karottenreibe:
rfr @salsolatragus 
This improves the local development workflow by
- making sure debug builds always build both variants of the profiler. This makes sure tests don't accidentally fail due to one variant not being built
- putting the output of the GUI and UploadDaemon next to the profiler binaries so we don't have to manually copy around files to be able to test the profiler on our machines

PR #79, opened 2018-12-20
Trigger upload on profiler shutdown.

### LINKED ISSUE #62: Profiler should trigger Upload on shutdown
Instead of uploading only by a time-based trigger, the profiler could launch the upload daemon again on shutdown. The upload daemon would then have to pass a message to any running daemon process to trigger an upload (after e.g. waiting a few seconds for safety).

This way, we could change the interval-based uploads to e.g. once per hour and on startup.

salsolatragus:
Fixes #62.

- [ ] Tests written
- [ ] Documentation updated

I prototyped a command forwarding that should allow us to control the `UploadDaemon` from the profiler. I'm unsure whether this solution covers all use cases we have and there's some decision I left open (TODOs in the code), because I lack information to decide what's the right approach. I would appreciate if we could have a talk about this, then it's probably straightforward to complete this.

### DIFF
--- a/UploadDaemon/UploadDaemon.cs
+++ b/UploadDaemon/UploadDaemon.cs
@@ -42,7 +42,7 @@ namespace UploadDaemon
         {
             if (IsAlreadyRunning())
             {
-                Console.WriteLine("Another instance is already running.");
+                logger.Debug("Another instance is already running. Sending notification to trigger upload.");
                 NotifyRunningDaemon();
                 return;
             }
--- a/UploadDaemon/UploadDaemon.cs
+++ b/UploadDaemon/UploadDaemon.cs
@@ -62,7 +62,7 @@ namespace UploadDaemon
 
             logger.Info("Starting upload daemon v{uploaderVersion}", Assembly.GetExecutingAssembly().GetName().Version.ToString());
             var uploader = new UploadDaemon(config);
-            uploader.RunOnce();
+            uploader.UploadOnce();
             uploader.ScheduleRegularUploads();
             uploader.WaitForNotifications();
         }
@@ -94,9 +94,9 @@ namespace UploadDaemon
         }
 
         /// <summary>
-        /// Runs the timer action once synchronously.
+        /// Uploads once, synchronously.
         /// </summary>
-        public void RunOnce()
+        public void UploadOnce()
         {
             lock(SequentialUploadsLock)
             {
@@ -110,7 +110,7 @@ namespace UploadDaemon
         private void ScheduleRegularUploads()
         {
             Timer timer = new Timer();
-            timer.Elapsed += (sender, args) => RunOnce();
+            timer.Elapsed += (sender, args) => UploadOnce();
             timer.Interval = TimerIntervalInMilliseconds;
             timer.Enabled = true;
         }
@@ -130,7 +130,7 @@ namespace UploadDaemon
                         // There is currently only one command (DaemonControlCommandUpload), hence,
                         // we immediately trigger an upload without checking what we received.
                         streamReader.ReadLine();
-                        RunOnce();
+                        UploadOnce();
                     }
                 }
             }
--- a/UploadDaemon/UploadDaemon.cs
+++ b/UploadDaemon/UploadDaemon.cs
@@ -1,7 +1,6 @@
 using Common;
 using NLog;
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Abstractions;
--- a/UploadDaemon/UploadDaemon.cs
+++ b/UploadDaemon/UploadDaemon.cs
@@ -127,24 +127,12 @@ namespace UploadDaemon
                     pipeServerStream.WaitForConnection();
                     using (var streamReader = new StreamReader(pipeServerStream))
                     {
-                        var command = streamReader.ReadLine();
-                        ExecuteCommand(command);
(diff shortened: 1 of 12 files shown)


salsolatragus:
@karottenreibe I've rebased this on your Yaml config PR and implemented it according to our discussion last week. Actually made things much simpler. Rfr.

UploadDaemon/UploadDaemon.cs
```diff
@@ -0,0 +1,168 @@
...
+        private readonly FileSystem fileSystem;
+
+        /// <summary>
+        /// Main entry point. Expects a single argument: the path to a directory that contains the trace files to upload.
+        /// </summary>
+        public static void Main(string[] args)
+        {
+            if (IsAlreadyRunning())
+            {
+                Console.WriteLine("Another instance is already running.");
```
karottenreibe: please also mention that an upload has been triggered

UploadDaemon/UploadDaemon.cs
```diff
@@ -0,0 +1,168 @@
...
+            if (config.DisableSslValidation)
+            {
+                HttpClientUtils.DisableSslValidation();
+            }
+
+            uploadTask = new UploadTask(config, fileSystem, new UploadFactory());
+        }
+
+        /// <summary>
+        /// Runs the timer action once synchronously.
```
karottenreibe: you renamed the timer action. please update the comment

UploadDaemon/UploadDaemon.cs
```diff
@@ -0,0 +1,168 @@
...
+        private void ScheduleRegularUploads()
+        {
+            Timer timer = new Timer();
+            timer.Elapsed += (sender, args) => RunOnce();
+            timer.Interval = TimerIntervalInMilliseconds;
+            timer.Enabled = true;
+        }
+
+        /// <summary>
+        /// Waits for notifications from subsequent executions of the Daemon. Terminates on
```
karottenreibe: where is that handling for the cancel key? do we really need that?

UploadDaemon/UploadDaemon.cs
```diff
@@ -0,0 +1,168 @@
...
+        private void WaitForNotifications()
+        {
+            while (true) // wait for indefinitely many commands
+            {
+                using (var pipeServerStream = new NamedPipeServerStream(DaemonControlPipeName, PipeDirection.In, 1, PipeTransmissionMode.Byte, PipeOptions.Asynchronous))
+                {
+                    pipeServerStream.WaitForConnection();
+                    using (var streamReader = new StreamReader(pipeServerStream))
+                    {
+                        var command = streamReader.ReadLine();
```
karottenreibe: do we have to trim or strip whitespace (i.e. the trailing \r\n) here?

UploadDaemon/UploadDaemon.cs
```diff
@@ -0,0 +1,168 @@
...
+                        ExecuteCommand(command);
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Executes a command forwarded to this UploadDaemon process.
+        /// </summary>
+        /// <param name="command"></param>
```
karottenreibe: please remove useless param tag

UploadDaemon/UploadDaemon.cs
```diff
@@ -0,0 +1,168 @@
...
+        private void ExecuteCommand(string command)
+        {
+            switch (command)
```
karottenreibe: why do we need this switch statement if we always do the same thing? remove? this looks very confusing. I'd prefer a line comment that there is currently only one command and thus we don't need to check the command. And then you can inline this method actually

UploadDaemon/UploadDaemon.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <summary>
+        /// Forwards a command to the existing UploadDaemon process.
+        /// </summary>
+        private static void NotifyRunningDaemon()
+        {
+            using (var pipeClientStream = new NamedPipeClientStream(".", DaemonControlPipeName, PipeDirection.Out, PipeOptions.Asynchronous))
+            {
+                pipeClientStream.Connect();
+
+                using (var w = new StreamWriter(pipeClientStream))
```
karottenreibe: please don't use one-letter variable names

UploadDaemon/UploadDaemon.cs
salsolatragus: I removed this already, after we agreed we don't need it on the phone. Updated the comment now as well. Good catch!

UploadDaemon/UploadDaemon.cs
salsolatragus: Nope, [`ReadLine()` does that for us](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.readline?redirectedfrom=MSDN&view=netframework-4.7.2#remarks).

salsolatragus:
rfr again

UploadDaemon/UploadDaemon.cs
```diff
@@ -42,7 +41,7 @@ public static void Main(string[] args)
         {
             if (IsAlreadyRunning())
             {
-                Console.WriteLine("Another instance is already running.");
+                logger.Debug("Another instance is already running. Sending notification to trigger upload.");
```
karottenreibe: this was written to the console on purpose: it's for the case where the user runs the executable from the command line. please change back to writeline. Logging with debug level will never be seen by anyone ;-)

PR #90, opened 2019-02-26
Bummer

### LINKED ISSUE #16: Use Bummer to translate trace files directly on the test machine
Should be optional. Finished traces should be processed with Bummer to receive line coverage. Output should be a coverage file that Teamscale can interpret (e.g. MS coverage or simple coverage format or whatever is the simplest to implement). This should be done by the Uploader component of the profiler.

It probably makes sense to run this translation as soon as the profiler shuts down to prevent someone from deleting the PDB files in between trace file writing and translation. Thus, the profiler should send a signal to the Uploader process when it's done. This can be done by e.g. launching the Uploader process again and the Uploader C# process can then communicate with any already running process. E.g. like shown here: https://stackoverflow.com/a/44828800/1396068

karottenreibe:
Fixes #16 

- [x] Tests written
- [x] Documentation updated


karottenreibe:
rfr

In addition to the automated tests I did a manual test against Pinta (open source drawing program) and an actual Teamscale instance. I checked both upload with timestamp and upload with revision. I also checked upload to a filesystem folder.

karottenreibe:
fixed merge conflicts

karottenreibe:
Fixed one more bug that happened during the latest pilot

karottenreibe:
merging as agreed with Flo

PR #100, opened 2019-03-06
use the correct flag to force a rejit in non-light-mode

### LINKED ISSUE #89: Evaluate whether use of COR_PRF_MONITOR_ENTERLEAVE should be replaced by COR_PRF_ENABLE_REJIT
We're [using](https://github.com/cqse/teamscale-profiler-dotnet/blob/master/Profiler/CProfilerCallback.cpp#L169) in the default (non-light) mode `COR_PRF_MONITOR_ENTERLEAVE` instead of `COR_PRF_ENABLE_REJIT`. According to the [documentation](https://docs.microsoft.com/de-de/dotnet/framework/unmanaged-api/profiling/cor-prf-monitor-enumeration) I'd expect that the latter.

Using the non-light mode caused problems at one of our customers when executing tests with the Tosca Classic DotNet engine (not TBox). Tosca cannot connect to the application due to a DLL injection error. Switching to light mode (and thus not monitoring enter/leave events) fixes that problem.

According to @karottenreibe we have to verify that exchanging the flags still produces correct output if the assembly is pre-jitted with `ngen.exe`. It would also be nice to have a performance comparison.

karottenreibe:
activating enter-leave callbacks is unnecessary and creates an
unnecessary performance overhead

Fixes #89 

- [ ] Tests written -> tested manually against Pinta
- [ ] Documentation updated -> not needed

### DIFF
--- a/Profiler/CProfilerCallback.cpp
+++ b/Profiler/CProfilerCallback.cpp
@@ -166,7 +166,7 @@ DWORD CProfilerCallback::getEventMask() {
 
 	// disable force re-jitting for the light variant
 	if (!config.shouldUseLightMode()) {
-		dwEventMask |= COR_PRF_MONITOR_ENTERLEAVE;
+		dwEventMask |= COR_PRF_DISABLE_ALL_NGEN_IMAGES;
 	}
 
 	return dwEventMask;


karottenreibe:
@mpdeimos do you want to review this? it's a one-liner ;-)

mpdeimos:
Thanks for investigating :+1: looks like my suspicion was right with the flag.

karottenreibe:
Definitely 😊 also cool that it was PR #100!

Martin Pöhlmann <notifications@github.com> schrieb am Mi., 6. März 2019,
16:14:

PR #107, opened 2019-03-19
Better defaults

### LINKED ISSUE #91: Make light mode and ignoring certificate errors the default
- light mode works in 99.9% of the cases and prevents performance problems - so enable it by default
- invalid or self-signed certificates are an often occurring error case during the uploader setup, so disable certificate validation by default

adjust the documentation accordinly

AnonymFx:
Fixes #91 

- [x] Tests written
- [x] Documentation updated

Note: I renamed "DisableSslValidation" to "EnableSslValidation" in the code to better reflect the default. I think it's more intuitive this way.

### DIFF
--- a/Common/ConfigParser.cs
+++ b/Common/ConfigParser.cs
@@ -21,9 +21,9 @@ namespace Common
             public List<ProcessSection> Match { get; set; }
 
             /// <summary>
-            /// Whether SSL validation should be globally enabled. Null if the user did not set this property.
+            /// Whether SSL validation should be globally disabled. Null if the user did not set this property.
             /// </summary>
-            public bool? EnableSslValidation { get; set; }
+            public bool? DisableSslValidation { get; set; }
         }
 
         /// <summary>
--- a/Common/ConfigParser.cs
+++ b/Common/ConfigParser.cs
@@ -21,9 +21,9 @@ namespace Common
             public List<ProcessSection> Match { get; set; }
 
             /// <summary>
-            /// Whether SSL validation should be globally disabled. Null if the user did not set this property.
+            /// Whether SSL validation should be globally enabled. Null if the user did not set this property.
             /// </summary>
-            public bool? DisableSslValidation { get; set; }
+            public bool? EnableSslValidation { get; set; }
         }
 
         /// <summary>
--- a/Profiler/Profiler.example.yml
+++ b/Profiler/Profiler.example.yml
@@ -7,7 +7,7 @@ match:
       enabled: false
       # Trace files will be stored and archived here
       targetdir: C:\Users\Public\Traces
-      # Light mode is recommended.
+      # Light mode is recommended and enabled by default.
       # Turn it off if you use ngen.exe for your assemblies
       light_mode: true
       # Enable upload to Teamscale
--- a/documentation/userguide.md
+++ b/documentation/userguide.md
@@ -408,11 +408,7 @@ off the proxy altogether and for using a different proxy.
 
 ## SSL certificates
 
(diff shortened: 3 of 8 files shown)


Common/ConfigParser.cs
```diff
@@ -21,9 +21,9 @@ public class YamlConfig
             public List<ProcessSection> Match { get; set; }
 
             /// <summary>
-            /// Whether SSL validation should be globally disabled. Null if the user did not set this property.
+            /// Whether SSL validation should be globally enabled. Null if the user did not set this property.
             /// </summary>
-            public bool? DisableSslValidation { get; set; }
+            public bool? EnableSslValidation { get; set; }
```
karottenreibe: This is a breaking change of the config format. We'll have to change all customer configs that are used right now. I'd like to avoid this and keep the name of the option as is if possible

Profiler/Profiler.example.yml
```diff
@@ -7,7 +7,7 @@ match:
       enabled: false
       # Trace files will be stored and archived here
       targetdir: C:\Users\Public\Traces
-      # Light mode is recommended.
+      # Light mode is recommended and enabled by default.
```
karottenreibe: Please add a comment here that light mode needs to be disabled in case the Native Image Cache is used

documentation/userguide.md
```diff
@@ -154,7 +154,7 @@ The profiler has several configuration options that can either be set as environ
 | :-------------------------------- | :--------------------------------------- | :--------------------------------------- |
 | COR_PROFILER_CONFIG               | Path                                     | Path to the profiler configuration file, e.g. `C:\Program Files\Coverage Profiler\profiler.yml` |
 | COR_PROFILER_TARGETDIR            | Path, default `c:/users/public/`         | Target directory for the trace files, e.g. `C:\Users\Public\Traces` |
-| COR_PROFILER_LIGHT_MODE           | `1` or `0`, default `0`                  | Enable ultra-light mode by disabling re-jitting of assemblies. |
+| COR_PROFILER_LIGHT_MODE           | `1` or `0`, default `1`                  | Enable ultra-light mode by disabling re-jitting of assemblies. |
```
karottenreibe: please amend the description with "Light mode must be disabled if you use the Native Image Cache"

documentation/userguide.md
```diff
@@ -408,11 +408,7 @@ off the proxy altogether and for using a different proxy.
 
 ## SSL certificates
 
-By default, the upload daemon will validate SSL certificates and refuse to upload to endpoints with invalid
-certificates. In case you must upload to an endpoint with a self-signed, certificate, you can use the
-config option `disableSslValidation` to ignore invalid certificates.
-
-This is insecure and not recommended.
+By default, the upload daemon will not validate SSL certificates. We strongly recommend turning on the validation with the `enableSslValidation` option in the config file. The reason why it's off by default is that, according to our experience, it needs to be turned off more often than not.
```
karottenreibe: I think the reason would be clearer like this: "It is disabled by default to allow for an easier setup in environments that use self-signed certificates or custom certificate authorities"

Common/ConfigParser.cs
AnonymFx: Good point, I didn't think of this. Then I'll change back all variable names in the code, so that we don't have both variants which actually adds confusion.

Profiler/Profiler.example.yml
AnonymFx: The line below this change already reflects this:  `Turn it off if you use ngen.exe for your assemblies`

karottenreibe: ah sorry, I hadn't seen that. my bad

karottenreibe:
good work!

PR #112, opened 2019-04-11
Attach log

AnonymFx:
Fixes #

- [x] Tests written
- [x] Documentation updated


AnonymFx:
I think the remaining 3 findings are fine. The destructors [here](https://demo.teamscale.com/findings.html#details/teamscale-net-profiler/?id=A59C2081930AEFBA4F839C0FBC342AB1&t=attach_log%3A1554988116000) and [here](https://demo.teamscale.com/findings.html#details/teamscale-net-profiler/?id=A730544859EC45C779FB3250AB5BACF9&t=attach_log%3A1554988116000) would just have empty bodies and the [missing interface comment](https://demo.teamscale.com/findings.html#details/teamscale-net-profiler/?id=B8DD6E370E5C6E06A648D9808D36691F&t=attach_log%3A1554988116000) is for a testing constant which should be clear without comment

karottenreibe:
unfortunately, I won't be able to do the review here before my vacation. @salsolatragus can you take over please?

Profiler/log/AttachLog.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+void AttachLog::createLogFile(std::string path) {
+	FileLogBase::createLogFile(path, "attach.log", false);
+}
+
+
+void AttachLog::logAttach() {
+	char timeStamp[BUFFER_SIZE];
+	getFormattedCurrentTime(timeStamp, sizeof(timeStamp));
+	std::string timeStampString(timeStamp);
```
salsolatragus: I would extract the logic to get a timestamp string into a helper method to remove duplication in `logDetach()`.

Profiler/log/FileLogBase.cpp
```diff
@@ -0,0 +1,110 @@
...
+FileLogBase::~FileLogBase()
+{
+	DeleteCriticalSection(&criticalSection);
+}
+
+void FileLogBase::createLogFile(std::string directory, std::string name, bool overwriteIfExists) {
+	if (directory.empty()) {
+		// c:\users\public is usually writable for everyone
+		// we must use backslashes here or the WinAPI path manipulation functions will fail
+		// to split the path correctly
```
salsolatragus: The comment is duplicated immediately after the `if`. Can we erase one copy?

Profiler/log/TraceLog.cpp
```diff
@@ -0,0 +1,58 @@
...
+void TraceLog::writeInlinedFunctionInfosToLog(std::vector<FunctionInfo>* functions)
+{
+	writeFunctionInfosToLog(LOG_KEY_INLINED, functions);
+}
+
+void TraceLog::createLogFile(Config& config) {
+	std::string targetDir = config.getTargetDir();
+
+	char timeStamp[BUFFER_SIZE];
+	getFormattedCurrentTime(timeStamp, sizeof(timeStamp));
```
salsolatragus: Here you have the logic for getting a timestamp string again (as in the `AttachLog`), but you don't need to convert it to a string before concatenating it into the log message. Why the difference? Can the same logic/helper method be used in all three places?

Profiler_Test/ProfilerTest.cs
```diff
@@ -128,5 +128,25 @@ public void TestProfiling(string application, int[] expectedAssemblyIds)
             List<FileInfo> traces = RunProfiler(application);
             AssertNormalizedTraceFileEqualsReference(traces, expectedAssemblyIds);
         }
+
+        [Test]
+        public void TestAttachLog()
+        {
+            RunProfiler("ProfilerTestee.exe", arguments: "all", lightMode: true, bitness: Bitness.x86);
+            string[] lines = File.ReadAllLines(AttachLog);
+            string lastAttachLine = lines[lines.Length - 2];
```
salsolatragus: The `-2` is pretty magic number to me. Can we somehow make this more self-explanatory? E.g. by extracting to 2 into a variable with a self-explaining name? It's the second last line you check here, right? Maybe renaming `lastAttachLine` to `secondLastLogLine` already helps? Why is it not simply the first line?

Profiler_Test/ProfilerTest.cs
```diff
@@ -128,5 +128,25 @@ public void TestProfiling(string application, int[] expectedAssemblyIds)
...
+            Assert.That(lastAttachLine.StartsWith("Attach"));
+            Assert.That(lastAttachLine.Contains("ProfilerTestee.exe"));
+        }
+
+        [Test]
+        public void TestDetatchLog()
+        {
+            RunProfiler("ProfilerTestee.exe", arguments: "all", lightMode: true, bitness: Bitness.x86);
+            string[] lines = File.ReadAllLines(AttachLog);
+            string lastDetachLine = lines.Last();
```
salsolatragus: `lastLogLine`?

documentation/userguide.md
```diff
@@ -244,6 +244,8 @@ Things to check if no trace files are written:
 In case the application doesn't start at all, please check the file `C:\Users\Public\profiler_debug.log`.
 It may contain stack traces in case the profiler crashed.
 
+The `attach.log` file in the directory of the profiler dlls might also provide some insights. It contains information about processes to which the profiler attached to.
```
salsolatragus: Join this paragraph and the previous, as they are both concerned with debugging startup failure. Please consistently capitalize "DLLs". The last "to" is too much, just say "to which the profiler attached".

Profiler/log/AttachLog.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+#include "FileLogBase.h"
+
+/**
+ * Manages a log file on the file system to which profiler attach events (time, executable name and process ID) are written.
+ * Unless mentioned otherwise, all methods in this class are thread-safe and perform their own synchronization.
+ */
+class AttachLog: public FileLogBase
```
salsolatragus: `AttachLog` is not really a `FileLogBase`... you don't use any of the methods declared by the supertype on the instances for `AttachLog`. I would suggest to implement `AttachLog` a separate and use a `FileLog` internally to implement it using delegation.

Profiler/CProfilerCallback.cpp
```diff
@@ -53,37 +53,42 @@ HRESULT CProfilerCallback::InitializeImplementation(IUnknown* pICorProfilerInfoU
 		return S_OK;
 	}
 
-	log.createLogFile(config);
-	log.info("looking for configuration options in: " + config.getConfigPath());
+	// Place the attach log next to the config and profiler dll
+	std::string configPath = StringUtils::removeLastPartOfPath(config.getConfigPath());
+	attachLog.createLogFile(configPath);
```
salsolatragus: I'm wondering whether we should write this log to the same directory as the trace log... We recently had a scenario where the profiler had to write permissions to the profiler install dir, such that logs couldn't be written there.

Profiler/CProfilerCallback.cpp
```diff
@@ -53,37 +53,42 @@ HRESULT CProfilerCallback::InitializeImplementation(IUnknown* pICorProfilerInfoU
...
 	}
 
-	log.createLogFile(config);
-	log.info("looking for configuration options in: " + config.getConfigPath());
+	// Place the attach log next to the config and profiler dll
+	std::string configPath = StringUtils::removeLastPartOfPath(config.getConfigPath());
+	attachLog.createLogFile(configPath);
+	attachLog.logAttach();
+
+	traceLog.createLogFile(config);
```
salsolatragus: For consistency, I think both `createLogFile` methods should take the directory path to write the log file to as an argument. Passing the entire `config` just to call `config.getTargetDir()` in the first line is unnecessary coupling anyways.

Profiler_Test/ProfilerTest.cs
AnonymFx: It's not the first line because the attach log is not newly created for the tests. If we used the first line of the file, we might get one that is pretty old (might even be days, weeks or longer, at least when testing on a dev machine). Since we want to test the log that we created with our current test run, we need to use the last and second last lines.

salsolatragus: Should we cleanup the files written by tests, say, in a `tearDown()`, to ensure there is no unexpected behavior due to data from previous tests or even test runs?

karottenreibe:
@AnonymFx this PR has been open for a pretty long time. Can you please have a look and discuss with @salsolatragus in case there's anything unclear/blocking you?

AnonymFx:
Unfortunately, it is time that is blocking me right now 😟

Profiler/log/AttachLog.h
AnonymFx: I'd rather move some of the log methods from `FileLogBase` to `TraceLog`. I think conceptionally `AttachLog` should be a `FileLogBase` and using delegation here would require a not insignificant amount of restructuring, which probably results in remove FileLogBase all together and moving it's functionality into Utils. Although possible, I'd prefer the solution mentioned at the start of the comment. What do you think?

Profiler/CProfilerCallback.cpp
AnonymFx: According to the [issue description](https://jira.cqse.eu/browse/TS-21153), not being able to write there is fine and should just result in no attach log being written at all.
I don't think it's necessarily a good idea to "pollute" the trace folder with something else than coverage files, but that's more a feeling than a real argument.

Profiler_Test/ProfilerTest.cs
AnonymFx: :+1:

AnonymFx:
Apologies for the huge delay, finally had time working on it again. Made some changes and commented the rest, so rfr again. @salsolatragus

Profiler/log/AttachLog.h
salsolatragus: I'd be ok with that, I guess :)

Profiler/CProfilerCallback.cpp
salsolatragus: Hm... I'm undecided about that one. I see your point regarding the pollution. I also see me totally confused, because none of the expected files are written and I missed that the process cannot write to the install dir xD

Writing the file in many cases is still ok and better than not having it at all. I'm fine to merge this and open another issue to change the path, if such confusion happens in practice.

salsolatragus:
Wow, this is really a long running PR 😅
Can we attach this to the Jira ticket and manage status there? Would make it easier to see the "in review"/"rework" states.

Profiler/CProfilerCallback.cpp
AnonymFx: Ok, then let's create a new ticket so that we can merge this as soon as possible. :)

Profiler/log/AttachLog.h
AnonymFx: Moved all log methods apart from `writeToFile` and `writeTupleToFile` which are needed in both attach and trace log.

AnonymFx:
Rfr again, set the status in Jira as well.

Profiler/CProfilerCallback.cpp
salsolatragus: Let’s mit create the ticket and wait for the issue to occur in practice, first.

salsolatragus:
Please have a look at the new Teamscale findings. Otherwise, I think we’re ready to merge.

PR #118, opened 2019-05-03
Batch upload

### LINKED ISSUE #110: Batch upload of coverage
When processing coverage, the uploader should batch uploads to Teamscale as much as possible to avoid having 100s of small uploads when one big one would do.

When implementing this, we will probably need to cache coverage on disk because multiple processes might be profiled wtih different partitions, revision.txt contents etc.

karottenreibe:
Fixes #110 

- [x] Tests written
- [x] Documentation updated


karottenreibe:
I tested the runtime of the merging once manually:

- creating a hashset with 100_000 files with 1_000 covered line ranges each takes about 35 seconds and uses about 6GB RAM according to the task manager
- 1_000_000 files with 100 covered ranges freezes my VirtualBox ;-) But this is not a realistic scenario anyways.
- adding a few more entries after that takes < 3ms as expected (HashSet insert is amortized O(1))
- on 32bit systems the above will not work since we run against the allocation limit of the CLR (System.OutOfMemoryException) even with less then 10_000 line ranges in one HashSet. So we can only merge large coverage files on 64bit systems. I'll add a flag to turn merging off in case we should ever need to

karottenreibe:
I checked the test cases you had written on your branch and I think I have the relevant ones covered on this branch now as well. If you agree, please delete your branch :-)

karottenreibe:
I checked the test cases you had written on your branch and I think I have the relevant ones covered on this branch now as well. If you agree, please delete your branch :-)

TS is currently reanalyzing since I deactivated a check that doesn't make sense for our code. I hope I fixed all findings

UploadDaemon/Upload/IUpload.cs
```diff
@@ -36,5 +36,14 @@ public interface IUpload
...
         string Describe();
+
+        /// <summary>
+        /// Returns an object that implements GetHashCode and Equals which can
+        /// be used as a key in a Dictionary that identifies this upload. I.e.
+        /// two keys from the same type of IUpload should only be equal if
+        /// they represent the same upload destination and coverage targeted
+        /// at both IUploads can be merged and uploaded in one go.
+        /// </summary>
+        object GetDictionaryKey();
```
salsolatragus: I don't like the technical name. Would `TargetId` or `DestinationId` describe the concept? I think if I would see we cluster by `upload.TargetId`, I would understand that we cluster by where the coverage goes to. Then we could shorten the comment to say that two `IUpload`s with the same target should return the same `TargetId`. I think the comment should not state what we use this for, because this might change. In fact, equal `TargetId`s are not sufficient for merging, since also the `Revision` must be the same. This key really only models the upload-target concept.

UploadDaemon_Test/MockUpload.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System.Threading.Tasks;
+using UploadDaemon.Upload;
+using UploadDaemon.SymbolAnalysis;
+
+/// <summary>
+/// Mock for the IUpload interface for testing.
+/// </summary>
+public class MockUpload : IUpload
```
salsolatragus: For the purpose of the merger tests this could be be replaced by the less verbose `Mock` instance.

```csharp
var upload1 = new Mock<IUpload>();
upload1.Setup(u => u.UploadAsync(any, any)).Return(Task.FromResult(true)); // * 
upload1.Setup(u => u. UploadLineCoverageAsync(any, any, any)).Return(Task.FromResult(true)); // *
upload1.Setup(u => u.GetDictionaryKey()).Return("key1");
```

In fact, the same would work for the `UploadTaskTests`, where the `LastUsedVersion` could be verified using

```csharp
upload1.Verify(u => u.UploadAsync(any, "expected version")); // *
```

*Not sure about the arg matcher syntax from the top of my head 😇

UploadDaemon_Test/UploadTaskTest.cs
```diff
@@ -222,62 +221,36 @@ public MockUploadFactory(bool returnValue)
...
+        /// <inheritDoc />
+        public Dictionary<string, FileCoverage> ConvertToLineCoverage(ParsedTraceFile traceFile, string symbolDirectory, GlobPatternList assemblyPatterns)
+        {
+            if (!shouldProduceCoverage)
+            {
+                return null;
+            }
+
+            FileCoverage fileCoverage = new FileCoverage();
+            fileCoverage.CoveredLineRanges.Add((12, 33));
```
salsolatragus: I don't like having these test expectations separate from the test themselves. Again, I would use a `Mock` in the tests and stub them such that they return what I need for the particular test. It's just a single method to stub, so it is done with a single `mockSynthesizer.Setup(synth => synth.ConverToLinge(arg, arg)).Return(x);` statement in each test.

documentation/userguide.md
```diff
@@ -415,6 +415,12 @@ By default, the upload daemon will not validate SSL certificates. We strongly re
 By default, the upload daemon will upload finished trace files every 5 minutes. To change this interval, set the `uploadIntervalInMinutes` option in the config file to the desired value.
 Set the option to `0` to disable scheduled uploads. The uploader will then upload finished trace files when invoked and terminate immediately after. This allows complete manual control of uploads.
 
+## Disabling Trace File Merging
```
salsolatragus: I would call this "Trace-File Merging" (because that's the feature).

documentation/userguide.md
```diff
@@ -415,6 +415,12 @@ By default, the upload daemon will not validate SSL certificates. We strongly re
...
+
+By default, the upload daemon will merge all line coverage that will be uploaded to the same
+destination into one file. This saves disk space and reduces the number of uploads.
+If this is not desired, you can turn this off by setting the `mergeLineCoverage` option to `false`.
```
salsolatragus: "...option in the config file to `false`"

UploadDaemon/SymbolAnalysis/FileCoverage.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// The line coverage collected for one file.
+    /// </summary>
+    public class FileCoverage
+    {
+        /// <summary>
+        /// The ranges of inclusive start and end lines that are covered in the file.
+        /// </summary>
+        public HashSet<(uint, uint)> CoveredLineRanges { get; } = new HashSet<(uint, uint)>();
+
+        public FileCoverage()
```
salsolatragus: Remove empty constructor.

salsolatragus:
Btw. the Teamscale check disappeared from the PR checks.

Common/TeamscaleServer.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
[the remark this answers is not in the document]
karottenreibe: It's sad that there's no auto-cleanup of these on save :-(

UploadDaemon/LineCoverageMerger.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// given revision/timestamp to the given upload to the merger. The coverage will be merged with
+        /// any existing coverage that should be uploaded to the same destination.
+        /// </summary>
+        public void AddLineCoverage(string traceFilePath, RevisionFileUtils.RevisionOrTimestamp revisionOrTimestamp, IUpload upload, Dictionary<string, FileCoverage> lineCoverage)
+        {
+            MergeKey key = new MergeKey
+            {
+                RevisionOrTimestamp = revisionOrTimestamp,
+                UploadDictionaryKey = upload.GetDictionaryKey(),
+                UploadType = upload.GetType()
```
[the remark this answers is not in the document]
karottenreibe: the latter, exactly. The line coverage files don't contain any meta data anymore so they are indistinguishable at that point.

UploadDaemon/Upload/IUpload.cs
karottenreibe: I agree with the rename but disagree with removing that part of the comment. Someone implementing a new type of upload needs to understand the implications of this method's implementation. "Target" is not specific enough to make someone understand what the implications for the line coverage merge process are. E.g. I think the intuitive implementation for the TS upload would be to just return all TeamcsaleServer fields, because that's the target right? But the username and access key should not factor into the line coverage merge decision, so that would not be correct. But that's not transparent to someone who doesn't know the merge logic (or even that it exists).
I also disagree that comments need to be generic enough so you'll never have to change them in the future, i.e. take every possible future use case into account. Comments - like any part of the source code - have to be maintained by us which includes adjusting them if they are no longer accurate. Just like identifiers have to be renamed when they no longer accurately reflect their underlying concepts.
Let's discuss in person if you are not convinced!

UploadDaemon_Test/MockUpload.cs
karottenreibe: While the mock is less verbose in the sense that it uses less lines, I find it harder to read.
Let's best discuss this in person

UploadDaemon_Test/UploadTaskTest.cs
karottenreibe: but the point of that is just to have _any_ coverage at all. The tests don't care which exact lines will be returned. Thus the only input to that class is true/false = return coverage or return null.

IMO this makes the tests much easier to read than having the same mock setup code in each test, which unnecessarily bloats the test case. I much prefer the tests to be as terse as possible so I can quickly figure out what it's doing and read details on demand if necessary

documentation/userguide.md
karottenreibe: AFAIK English doesn't use a hyphen here. But I'm not 100% sure if you meant "insert a hyphen between trace and file" with your comment or "change the headline to 'trace file merging' instead of 'disabling trace file merging'"?

UploadDaemon/SymbolAnalysis/FileCoverage.cs
karottenreibe: cool! I thought I needed that ;-) good catch

karottenreibe:
I think the TS problem is due to me reanalyzing the project last week. It's back now after pushing a new commit

karottenreibe:
rfr again. Please see my answers to several of your comments above in the GitHub discussion. Can you call me today so we can discuss in person? I have a tight schedule but maybe during lunch or coffee break?

Common/TeamscaleServer.cs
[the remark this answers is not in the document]
salsolatragus: I have the Productivity Power Tools installed, which provide this functionality :)

UploadDaemon/LineCoverageMerger.cs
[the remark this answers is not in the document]
salsolatragus: Ok, then I'm fine with leaving it as is. Maybe document this reasoning?

karottenreibe:
will resolve the conflicts now

UploadDaemon/Upload/IUpload.cs
salsolatragus: If you have doubts that somebody else would infer the right consequences from the name alone, I'm fine with leaving the additional information. Better to err on the side of too much details, I guess.

I absolutely agree that comments need to be maintained. Unfortunately, in my experience, this is more easily forgotten than updating code, because an outdated comment doesn't (directly) break any functionality. Therefore, I try to place the comments close to where I would need to do code changes that would require to update the comment, such that I'm likely to see them when I change the code and they are likely to appear in the review diff. Therefore, I'm concerned about a getter that documents how the return value is used. It's easy to change the usage (or add new, different usages) without every seeing/updating the comment. I would prefer to document the properties of the return value, independent of the usage.

How about something like: "Two `IUpload` should return equal target Ids if, and only if, the result of uploading coverage through either of them will be the same, i.e., the coverage will be uploaded to the same destination and it will be indistinguishable which of the uploads was used, after the fact."

salsolatragus: Sorry for writing this out instead of calling you, it was easier to formulate a draft in typing 😇

karottenreibe: Sounds good, I like it

karottenreibe:
implemented your suggestions from the comments above. We still need to discuss the test code

documentation/userguide.md
salsolatragus: Both, actually ;)

I think the section should describe the feature, including how to disable it. As a user, I want the guide to tell me what the tool can do. Therefore, I would remove the "Disabling" in the header.

The hyphen is required at this point, because otherwise it would mean that there is a file merging that you trace, as opposed to a trace file that your merge. This is because [english is right-associative](https://www.bodden.de/2008/07/16/english-is-right-associative-french-is-not/).

salsolatragus:
Ok, then, let's have a call about how we mock in our tests :) Since you have the tight schedule, just call me whenever it fits. I have no appointments until 5pm.

PR #119, opened 2019-05-03
Garbage Collect Trace Files

### LINKED ISSUE #114: Garbage Collection of coverage files
We currently archive coverage files after upload. However, this may imply huge amounts of data accumulate over time. We should have some garbage collection routine (potentially in the UploadDaemon, since only then new files are added to the archives) that delete "old" coverage files. Age could be defined by the number of days since the files was added and "old" should then be configurable as a number of days as well.

salsolatragus:
Fixes #114

- [x] Tests written
- [x] Documentation updated


karottenreibe:
@salsolatragus You seem to have changed the line endings in several files which messes up the diff. Can you somehow correct this and check why this happened? Otherwise the review is really painful since I have to manually compare each line of those files :-(

salsolatragus:
Somehow my VS sometimes decides to change the line endings because they are, apparently, inconsistent. I don’t see this in my diff tool, because it’s ignoring like endings... I think you can achieve the same on GitHub. There‘s an option to ignore white space somewhere on the review bar at the top. Does that work for you? I’m on the bus home and cannot commit till later tonight.

karottenreibe:
Ah thanks, didn't know that!
With the line GitHub GUI I can now diff at least the files that remained in the same location. Moved files are still shown as fully deleted and fully added but I'll muddle through.

For the future: I think you can configure your git to always check in with a fixed set of line endings. There are some magic options for that. Then it doesn't matter what VS does.

karottenreibe:
I run VS in a VirtualBox but did the git checkout from my linux host and check in via that as well which seems to auto-fix the line endings to linux style. Maybe that works for your setup as well?

UploadDaemon_Test/Config/ConfigTest.cs
```diff
@@ -408,7 +408,39 @@ public void TestConfigureUploadInterval()
...
+        [Test]
+        public void TestConfigurePurgeThresholdDefaults()
+        {
+            Config config = Config.Read(@"
+                match:
+                    - profiler:
+                        targetdir: C:\test1
+            ");
+
+            Assert.That(config.ArchivePurgingThresholds.UploadedTraces, Is.LessThan(TimeSpan.Zero));
```
karottenreibe: please add a test where only some of the options are specified and verify that the missing options get the correct  default value

Common/ConfigParser.cs
```diff
@@ -127,6 +133,27 @@ public class UploaderSubsection
...
 
+        /// <summary>
+        /// Contains the thresholds for purging trace files from the upload archives.
+        /// </summary>
+        public class PurgeUploadArchivesSection
+        {
+            /// <summary>
+            /// Minimal timespan to keep uploaded traces.
+            /// </summary>
+            public TimeSpan UploadedTraces { get; set; } = TimeSpan.MinValue;
```
karottenreibe: After reading the rest of the code I understand that MinValue here causes the uploader to not perform a purge. But when I initially saw this line, I thought that the default is "clean immediately". Can we represent the default in a cleaner way somehow to make it clear to the reader here that the default is "do nothing" and not "clean up everything right away"?

Also, the documentation of TimeSpan.MinValue never mentions that this is a negative value, so even after reading that I didn't understand this default value. Since we already use TimeSpan.Zero to mean "clean up right away", how about using null to represent "do nothing" and documenting that in the comment above the fields?

karottenreibe: For the UploadInterval you used an int for parsing, here you use TimeSpans. Is this on purpose? I'd expect us to use the same type for both.

Does the YAML parser guarantee that ints will always be deserialized into timespans with unit "Days" or is this more or less by accident? In fact, looking over the tests that we have, I don't think we actually verify that. I'd expect that this means that this is a bug because the Timespan will be initialized with the given number but that will be interpreted as "ticks" not as "days". So I'd prefer that we use ints for deserializing and convert to timespans ourselves when we construct the Config from the YamlConfig. That way we should be on the safe side :-)

salsolatragus:
The config option is `git config --global core.autocrlf input` (on Linux or OSX), which automatically fixes all line endings to *nix style on commit.

Common/ConfigParser.cs
salsolatragus: This is not by accident. Json.NET explicitly support [(de)serializing TimeSpans](https://www.newtonsoft.com/json/help/html/SerializeDateTimeZoneHandling.htm). The full format for a TimeSpan is "1.00:00:00.000" (d.hh:mm:ss.mmm). For days the abbreviation to an int is accepted by the parser. For minutes we would could write "00:05:00" in the config, but I didn't like that.

Correct conversion is tested: two tests in `ConfigTest` expect days timespans according to the ints in the config file.

salsolatragus:
rfr again

karottenreibe:
the TS findings are false-positives. Can you please blacklist those so the next time we edit the code we don't have to look at them again?

salsolatragus:
Did the changes as suggested. Unfortunately, I cannot blacklist the TS findings, because I don't seem to have the necessary privileges on the instance...

karottenreibe:
can you please check again in TS? I created a "CQSE" group with project admin rights, so you should now be able to do everything for that project :-)

karottenreibe:
great work!

salsolatragus:
@karottenreibe Yes, I can blacklist things now.

PR #126, opened 2019-05-16
Fix crashes on Windows 7

### LINKED ISSUE #125: Profiler crashes apps on Windows 7
When I start an app on Windows 7 with the profiler attached, app startup fails with `CLR error: 80004005. The program will now terminate`. The profiler is loaded into the runtime correctly, but at somepoint it crashes and takes the runtime down with it. The error seems to happen at different places... one we caught was:

```
KERNELBASE.dll!000007fefcc21b54()	Unknown
kernel32.dll!0000000076c91ce6()	Unknown
Profiler64.dll!Debug::logInternal(std::basic_string<char,std::char_traits<char>,std::allocator<char> > message) Line 40	C++
Profiler64.dll!Debug::log(std::basic_string<char,std::char_traits<char>,std::allocator<char> > message) Line 15	C++
Profiler64.dll!Debug::logStacktrace(std::basic_string<char,std::char_traits<char>,std::allocator<char> > context) Line 44	C++
Profiler64.dll!CProfilerCallback::handleException(std::basic_string<char,std::char_traits<char>,std::allocator<char> > context) Line 302	C++
Profiler64.dll!`CProfilerCallback::Shutdown'::`1'::catch$1() Line 157	C++
[External Code]	
Profiler64.dll!CProfilerCallback::Shutdown() Line 154	C++
Profiler64.dll!DllMain(HINSTANCE__ * hInstance, unsigned long dwReason, void * lpReserved) Line 23	C++
[External Code]
```

The error that brings us down is an [access violation](https://stackoverflow.com/questions/17168982/exception-error-c0000005-in-vc). However, this is not the original error (see `handleException()` on the stack. We couldn't capture the original error yet.

salsolatragus:
Fixes #125 

Turns out the problem was that we pass NULL to the "number of characters written" parameter of `WriteFile`, which is apparently valid under Win10, but not Win7. We now pass a reference and, voila, all works fine under Win7 again as well.

Unfortunately, AppVeyor doesn't seem to provide an image for running CI on Win7 as well...

No Teamscale findings.

### DIFF
--- a/Profiler/CProfilerCallback.cpp
+++ b/Profiler/CProfilerCallback.cpp
@@ -298,7 +298,7 @@ HRESULT CProfilerCallback::JITCompilationFinished(FunctionID functionId,
 }
 
 void CProfilerCallback::handleException(std::string context) {
-	debugLog.logStacktrace(context);
+	debugLog.logErrorWithStracktrace(context);
 	if (!config.shouldIgnoreExceptions()) {
 		throw;
 	}
--- a/Profiler/CProfilerCallback.cpp
+++ b/Profiler/CProfilerCallback.cpp
@@ -123,7 +123,7 @@ void CProfilerCallback::initializeConfig() {
 	if (!configFileWasManuallySpecified) {
 		configFile = Config::getDefaultConfigPath();
 	}
-	Debug::log(configFile);
+	//Debug::log(configFile);
 
 	config.load(configFile, WindowsUtils::getPathOfThisProcess(), configFileWasManuallySpecified);
 }
@@ -181,7 +181,7 @@ UINT_PTR CProfilerCallback::functionMapper(FunctionID functionId, BOOL* pbHookFu
 		return functionId;
 	}
 	catch (...) {
-		Debug::logStacktrace("functionMapper");
+		//Debug::logStacktrace("functionMapper");
 		// since this function must be static, we have no way to access the config so we always terminate the program.
 		throw;
 	}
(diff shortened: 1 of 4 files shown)


Profiler/CProfilerCallback.cpp
```diff
@@ -3,7 +3,6 @@
 #include "UploadDaemon.h"
 #include "utils/StringUtils.h"
 #include "utils/WindowsUtils.h"
-#include "utils/Debug.h"
```
karottenreibe: I'd prefer not to make this part of the classes interface. See also my comment at the end about the singleton pattern

Profiler/utils/Debug.h
```diff
@@ -9,32 +9,16 @@
...
-	Debug();
-
-	void logInternal(std::string message);
-
-	/**
-	 * Implements a thread-safe singleton pattern and ensures that the instance
-	 * is properly destroyed and thus the file handle freed.
-	 * See https://stackoverflow.com/a/1008289/1396068
-	 */
-	static Debug& getInstance()
```
karottenreibe: The idea behind making this a singleton was that you can call the log from any part of our code (e.g. utils functions) and have a thread-safe log output. Right now, I either have to explicitly pass the debugLog instance around in the code if I want to debug something or I have to create a new instance, which is not thread safe, i.e. it's possible to get mangled log output with overlapping log messages. I'd prefer to have this as a singleton again

Profiler/utils/Debug.h
```diff
@@ -9,32 +9,16 @@
 class Debug
 {
 public:
+	Debug();
```
karottenreibe: If you agree with my comment about this being a singleton, please make constructor and destructor private

salsolatragus:
This is part of #131. Closing this PR.

PR #129, opened 2019-05-28
123 archive simple coverage

### LINKED ISSUE #123: Introduce option to archive merged coverage instead of original files
Currently, when we merge line coverage into a single file, we still archive the original files locally after upload. It would be nice if we had the option to, instead, archive the merged file, because it is likely to be much smaller and should be equivalent in terms of the contained information.

karottenreibe:
Fixes #123

- [x] Tests written
- [x] Documentation updated


karottenreibe:
rfr @salsolatragus

PR #133, opened 2019-08-16
enforce usage of TLS 1.2

### LINKED ISSUE #137: Allow using TLS 1.2 on older .NET versions
.NET < 4.6.1 seems to not allow the use of TLS1.2. On the other hand, Teamscale seems to be configured to require it since older versions offer less security. Thus, on these .NET versions, the UploadDaemon and Teamscale can't communicate via HTTPS, leading to hard to debug problems.

We should explicitly enable TLS1.2.

If possible, the profiler should support also older protocols to support customers that still have old network stacks that require us to talk e.g. SSL3 or other deprecated protocols.

karottenreibe:
Fixes #137 

Teamscale requires TLS 1.2 but .NET framework < 4.6.1 does not offer that by default. We need to explicitly offer it. See e.g. https://blogs.perficient.com/2016/04/28/tsl-1-2-and-net-support/ or https://stackoverflow.com/a/37869237/1396068

- [ ] Tests written - no, unclear how this could be tested automatically
- [x] Documentation updated - not needed


karottenreibe:
@salsolatragus this is most likely the same bug we had with that other customer and Win 7 where we enabled weak ciphers in Teamscale :-)

karottenreibe:
@mpdeimos do you have time to review this?

karottenreibe:
@AnonymFx: Could you please take over from @mpdeimos here as well? It's only a few changed lines :-)

AnonymFx:
Sure

salsolatragus:
@AnonymFx Is there any reason why this is not merged yet?

karottenreibe:
@salsolatragus I think Pete had too little time for this topic. If you'd like, you can of course take over the review :-) (but please sync with Pete to make sure you don't do the work twice!)

AnonymFx:
I actually looked over the changes and approved them, just didn't remember that is was my job to merge. Was out of the dev business for a while after all 😅 @karottenreibe @salsolatragus

AnonymFx:
Merging now ;)

salsolatragus:
That why I brought it up ;)

salsolatragus:
Thank!

PR #151, opened 2019-12-18
CR#21447 Slow PDB Resolution

salsolatragus:
Fixes issue TS-21447

- [x] Tests written
- [ ] ~Documentation updated~ (performance and logging improvements only)
- [x] `CHANGELOG.md` updated


salsolatragus:
Requesting review from you, @karottenreibe, because I think you are the only one who can confirm that my caching strategy works correctly. It seems to work for the concrete pilot use case at hand, so merging this is not urgent.

UploadDaemon/Configuration/GlobPatternList.cs
```diff
@@ -75,5 +75,17 @@ public Boolean Matches(string text)
         {
             return includeRegexes.Any(regex => regex.IsMatch(text)) && !excludeRegexes.Any(regex => regex.IsMatch(text));
         }
+
+        /// <inheritdoc/>
+        public override bool Equals(object obj)
+        {
+            return this.Describe().Equals((obj as GlobPatternList)?.Describe());
```
karottenreibe: that seems hacky. I wouldn't expect that changing the describe method influences equals, so this is a chance for very subtle bugs. Can you please change to a proper eqauls method?

UploadDaemon/Configuration/GlobPatternList.cs
```diff
@@ -75,5 +75,17 @@ public Boolean Matches(string text)
...
+        }
+
+        /// <inheritdoc/>
+        public override int GetHashCode()
+        {
+            return this.Describe().GetHashCode();
```
karottenreibe: same here, please use a proper hash code. Martin P showed me a cool trick to make hashcode short in C#. C.f. here https://github.com/cqse/teamscale-profiler-dotnet/blob/f3ece7745aae386c87ffbcd4dab3d8b9a2a019b1/UploadDaemon/Configuration/AzureFileStorage.cs

UploadDaemon/SymbolAnalysis/LineCoverageSynthesizer.cs
```diff
@@ -15,10 +15,13 @@ public class LineCoverageSynthesizer : ILineCoverageSynthesizer
     {
         private static readonly Logger logger = LogManager.GetCurrentClassLogger();
 
+        private readonly IDictionary<string, IDictionary<GlobPatternList, SymbolCollection>> symbolCollectionsCache = new Dictionary<string, IDictionary<GlobPatternList, SymbolCollection>>();
```
karottenreibe: this data structure is very hard to comprehend. If I get this correctly, you want to use the symbol directory and the glob pattern list as a composite key to index the cached symbol collections, right?
For the sake of readability, I'd make this explicit: I think you can use tuples as Dictionary keys, which would a) make the intent clearer and b) simplify the code to access the dictionary.
Also, I'd move the cache into a separate class and use it from the SymbolCollection instead of from here. It seems to me the synthesizer shouldn't care whether it is using fresh or cached symbols. It just wants the symbols. So the Collection can handle the caching transparently
Making it its own class also allows us to add cache invalidation. Because one problem I see with this implementation is: if they redeploy their application and pdbs but don't kill the upload daemon, the deamon will incorrectly use the cached, outdated symbols against the new trace files. This will result in errors and incorrect coverage. We could e.g. in addition to the symbols cache some meta info about the pdbs (e.g. file names, last modified dates, file sizes?) and invalidate the cache if these don't match what we find on disk. Of course, we'd have to time how long gathering the meta data takes - would be bad if that's the bottleneck and not parsing the pdbs. Maybe some metadata can be read rather fast and some is slow (e.g. file size?)

salsolatragus: You got my intent right :)

I switched to using a `Tuple` and moved the caching into a dedicated `Resolver` class. Thanks for these suggestions.

I'm unsure about the invalidation. Could we somehow achieve invalidation by looking at the traces? The original issue was that reading the PDB files from a network share is slow. I fear if we introduce any reading from this source with every resolution, performance would regress here...

karottenreibe: I don't see how. The traces know nothing about PDBs. If they really have such a slow network share that listing the modified date of a few dozen files takes too long, I'd suggest not storing the PDBs on the network share ;-)
Could add a "don't invalidate pdb cache" option that is "use at your own risk". The default behaviour IMO **must** be that we properly invalidate the cache as otherwise we get very hard to debug problems with incorrect coverage. I don't want to go down that road ;-)

UploadDaemon/Configuration/GlobPatternList.cs
```diff
@@ -77,15 +77,10 @@ public Boolean Matches(string text)
         }
 
         /// <inheritdoc/>
-        public override bool Equals(object obj)
-        {
-            return this.Describe().Equals((obj as GlobPatternList)?.Describe());
-        }
+        public override bool Equals(object other) => other is GlobPatternList gpl && gpl.includePatterns.SequenceEqual(includePatterns)
```
karottenreibe: Please don't use abbreviations for variables (`gpl`) how about `list` or `otherList` instead?

UploadDaemon/SymbolAnalysis/SymbolCollectionResolver.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// This resolver loads symbol collections (PDB files) into <see cref="SymbolCollection"/>s.
+    /// It uses a cache to avoid unnecessary I/O from reading the same PDB files repeatedly.
+    /// </summary>
+    public class SymbolCollectionResolver
+    {
+        private readonly IDictionary<Tuple<string, GlobPatternList>, SymbolCollection> symbolCollectionsCache =
+            new Dictionary<Tuple<string, GlobPatternList>, SymbolCollection>();
+
+        public SymbolCollection ResolveFrom(string symbolDirectory, GlobPatternList assemblyPatterns)
+        {
```
karottenreibe: Please check the TS findings :-)

UploadDaemon/SymbolAnalysis/SymbolCollectionResolver.cs
```diff
@@ -0,0 +1,31 @@
...
+            if (!symbolCollectionsCache.ContainsKey(key))
+            {
+                symbolCollectionsCache[key] = SymbolCollection.CreateFromPdbFiles(symbolDirectory, assemblyPatterns);
+            }
+            return symbolCollectionsCache[key];
+        }
+
+        private static Tuple<string, GlobPatternList> GetCacheKey(string symbolDirectory, GlobPatternList assemblyPatterns)
+        {
+            return (symbolDirectory, assemblyPatterns).ToTuple();
```
karottenreibe: calling ToTuple on a tuple seems kind of redundant. Can't you just change the return type to `(string, GlobPatternList)` and use the same type in the map? Or is there a technical reason why that won't work?

UploadDaemon/SymbolAnalysis/LineCoverageSynthesizer.cs
```diff
@@ -15,12 +15,17 @@ public class LineCoverageSynthesizer : ILineCoverageSynthesizer
     {
         private static readonly Logger logger = LogManager.GetCurrentClassLogger();
 
-        private readonly IDictionary<string, IDictionary<GlobPatternList, SymbolCollection>> symbolCollectionsCache = new Dictionary<string, IDictionary<GlobPatternList, SymbolCollection>>();
+        private readonly SymbolCollectionResolver symbolCollectionResolver;
+
+        public LineCoverageSynthesizer(SymbolCollectionResolver symbolCollectionResolver)
```
karottenreibe: it seems that at every location where the `LineCoverageSynthesizer` is instantiated, you simply pass `new SymbolCollectionResolver()` here. If there's no reason at the moment to pass anything else, I'd say you can inline this in the field definiiton and thus make constructing this class a bit simpler.

UploadDaemon/SymbolAnalysis/SymbolCollectionResolver.cs
salsolatragus: The problem is that `(symbolCOllection, assemblyPatterns)` is, in fact, not a `Tuple`. Using `(string, GlobalPatternList)` works, though. I wasn't aware that this is a valid type (declaration).

UploadDaemon/SymbolAnalysis/LineCoverageSynthesizer.cs
salsolatragus: Implemented cache invalidation based on the `LastWriteDate`. Struggled somewhat to write a test for this that runs both on my machine and on AppVeyor (because setting the `LastWriteDate` has some funny side effects, such as setting the date to `Now` if called multiple times consequently...), but it seems I succeeded :)

UploadDaemon/SymbolAnalysis/SymbolCollectionResolver.cs
```diff
@@ -1,35 +1,59 @@
...
         }
 
         private static (string, GlobPatternList) GetCacheKey(string symbolDirectory, GlobPatternList assemblyPatterns)
         {
             return (symbolDirectory, assemblyPatterns);
         }
+
+        private bool IsValid(SymbolCollection collection)
+        {
+            return collection.SymbolFileNames.All(symbolFileName => ReadLastWriteDate(symbolFileName).Equals(symbolFileLastWriteDates[symbolFileName]));
```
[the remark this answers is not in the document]
karottenreibe: also, will this throw if the file no longer exists? we should handle that exception in either case and not terminate the daemon :-)

UploadDaemon/SymbolAnalysis/SymbolCollection.cs
```diff
@@ -70,6 +72,11 @@ public SymbolCollection(List<AssemblyMethodMappings> mappings)
         /// </summary>
         public bool IsEmpty => mappings.Count() == 0 || mappings.Values.All(mapping => mapping.Count() == 0);
 
+        /// <summary>
+        /// The paths of the PDB symbol files considered in this collection.
+        /// </summary>
+        public ICollection<string> SymbolFileNames { get; private set; }
```
karottenreibe, line 78: Rename to SymbolFilePaths? Name is confusing for me as I expect this to be just the file name

UploadDaemon_Test/SymbolAnalysis/SymbolCollectionResolverTest.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.IO;
+using System.Linq;
+using UploadDaemon.Configuration;
+
+namespace UploadDaemon.SymbolAnalysis
+{
+    [TestFixture]
+    class SymbolCollectionResolverTest
+    {
+        private readonly GlobPatternList someAssemblyPatterns = new GlobPatternList(new List<string> { "*" }, new List<string> { });
```
karottenreibe, line 13: the name is confusing for me. How about `includeAllAssembliesPattern`? Reflects better what it does and makes the tests easier to read

UploadDaemon_Test/SymbolAnalysis/SymbolCollectionResolverTest.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            SimulateSymbolFileChange(collection1.SymbolFileNames.First());
+
+            SymbolCollection collection2 = resolver.ResolveFrom(TestUtils.TestDataDirectory, someAssemblyPatterns);
+
+            Assert.That(collection1, Is.Not.SameAs(collection2));
+        }
+
+        [Test]
+        public void CachesCollectionAgainAfterInvalidation()
```
karottenreibe, line 55: please add a test for a file deletion

salsolatragus: Added tests for new PDBs and deleting PDBs. Deletion actually worked already. New files now work as well.

UploadDaemon/SymbolAnalysis/SymbolCollectionResolver.cs
```diff
@@ -41,19 +52,32 @@ private static (string, GlobPatternList) GetCacheKey(string symbolDirectory, Glo
...
-            return collection.SymbolFileNames.All(symbolFileName => ReadLastWriteDate(symbolFileName).Equals(symbolFileLastWriteDates[symbolFileName]));
+            return ContainsExactly(collection, relevantSymbolFiles)
+                && relevantSymbolFiles.All(symbolFile => symbolFile.LastWriteTime.Equals(symbolFileLastWriteDates[symbolFile.Path]));
         }
 
-        private void UpdateValidationInfo(SymbolCollection collection)
+        private static bool ContainsExactly(SymbolCollection collection, ICollection<SymbolFileInfo> relevantSymbolFilePaths)
         {
-            foreach (string symbolFileName in collection.SymbolFileNames)
+            return new HashSet<string>(relevantSymbolFilePaths.Select(info => info.Path)).SetEquals(collection.SymbolFilePaths);
```
karottenreibe, line 63: Minor: AFAICS, symbolfilepaths is only used for this check. Could we make that a set, so that we don't need to create a hashset here every time? would make the code easier to read and also make it clearer what the purpose of the SymbolFilePaths set is :-)

UploadDaemon/SymbolAnalysis/SymbolCollectionResolver.cs
```diff
@@ -41,19 +52,32 @@ private static (string, GlobPatternList) GetCacheKey(string symbolDirectory, Glo
...
 
-        private static DateTime ReadLastWriteDate(string symbolFileName) => File.GetLastWriteTimeUtc(symbolFileName);
+        private static ICollection<SymbolFileInfo> FindRelevantSymbolFiles(string symbolDirectory, GlobPatternList assemblyPatterns)
+        {
+            return new HashSet<SymbolFileInfo>(Directory.EnumerateFiles(symbolDirectory, "*.pdb", SearchOption.AllDirectories)
+                .Where(file => assemblyPatterns.Matches(Path.GetFileNameWithoutExtension(file)))
+                .Select(file => new SymbolFileInfo { Path = file, LastWriteTime = ReadLastWriteDate(file) }));
+        }
+
+        private static DateTime ReadLastWriteDate(string symbolFilePath) => File.GetLastWriteTimeUtc(symbolFilePath);
```
karottenreibe, line 81: only used onnce now. inline?

UploadDaemon_Test/SymbolAnalysis/SymbolCollectionResolverTest.cs
```diff
@@ -10,56 +10,113 @@ namespace UploadDaemon.SymbolAnalysis
     [TestFixture]
     class SymbolCollectionResolverTest
     {
-        private readonly GlobPatternList someAssemblyPatterns = new GlobPatternList(new List<string> { "*" }, new List<string> { });
-        private readonly string testSymbolFileName = $"{TestUtils.TestDataDirectory}\\Some.pdb";
+        private static readonly string TestDataDirectory = $"{TestUtils.TestDataDirectory}\\Tmp";
```
karottenreibe, line 13: I'd prefer using the system temp dir if possible. otherwise you may get a dirty workspace if running the tests aborts, which is annoying during local development

karottenreibe: if that's not easily possible with .NET, however, it's OK to leave it as is

karottenreibe: however, i'd rename it to `TestTempDirectory` to make the difference between this and `TestUtils.TestDataDirectory` more apparent. it's easy to confuse the two otherwise

UploadDaemon_Test/SymbolAnalysis/SymbolCollectionResolverTest.cs
```diff
@@ -10,56 +10,113 @@ namespace UploadDaemon.SymbolAnalysis
...
+            Directory.Delete(TestDataDirectory, true);
+        }
+
+        [Test]
+        public void ConsidersSymbolFileIncludes()
+        {
+            SymbolCollection collection = resolver.ResolveFrom(TestUtils.TestDataDirectory,
+                   new GlobPatternList(new List<string> { "DoesNotExist*" }, new List<string> {}));
+
+            Assert.That(collection.IsEmpty, Is.True);
```
karottenreibe, line 42: Minor: Is there a `Is.Empty` for collections? will probably give you better error messages. same for the next test

UploadDaemon_Test/SymbolAnalysis/SymbolCollectionResolverTest.cs
```diff
@@ -10,56 +10,113 @@ namespace UploadDaemon.SymbolAnalysis
     [TestFixture]
     class SymbolCollectionResolverTest
     {
-        private readonly GlobPatternList someAssemblyPatterns = new GlobPatternList(new List<string> { "*" }, new List<string> { });
-        private readonly string testSymbolFileName = $"{TestUtils.TestDataDirectory}\\Some.pdb";
+        private static readonly string TestDataDirectory = $"{TestUtils.TestDataDirectory}\\Tmp";
+        private static readonly string TestSymbolFilePath = $"{TestDataDirectory}\\Some.pdb";
+
+        private readonly GlobPatternList includeAllAssembliesPattern = new GlobPatternList(new List<string> { "*" }, new List<string> { });
```
karottenreibe, line 16: make static?

UploadDaemon_Test/SymbolAnalysis/SymbolCollectionResolverTest.cs
salsolatragus: Yes, there is, but `SymbolCollection` is not a .NET collection and, therefore, the `Is.Empty` matcher doesn't work.

karottenreibe: makes sense :-)

PR #152, opened 2020-08-04
CR#24019 Add eagerness to example config

AnonymFx:
Addresses issue [TS-24019](https://jira.cqse.eu/browse/TS-24019)

- ~~[ ] Tests written~~
- ~~[ ] Documentation updated~~
- ~~[ ] `CHANGELOG.md` updated~~
- ~~[ ] Present new features in [N&N](https://wiki.cqse.eu/pages/viewpage.action?pageId=689566)~~

Strikethroughs not required, only small example config change


Profiler/Profiler.example.yml
```diff
@@ -23,6 +23,10 @@ match:
       # Light mode is recommended and enabled by default.
       # Turn it off if you use ngen.exe for your assemblies
       light_mode: true
+      # Enable eager writing of traces after the specified amount
+      # of method calls (i.e. write to disk immediately).
+      # This should only be used in conjunction with light mode.
+      eagerness: 0
```
karottenreibe, line 29: Can you please add a use case, so users know when this should be used? IMO the only valid use case is if the profiler is killed before it can finish writing the trace files

PR #155, opened 2021-03-16
Partial Reports & Improved Merging

mpdeimos:


UploadDaemon/Report/Testwise/Test.cs
```diff
@@ -75,7 +75,12 @@ public Test Union(Test other)
 
         public SimpleCoverageReport ToSimpleCoverageReport()
         {
-            return new SimpleCoverageReport(CoverageByPath.First().Files.ToDictionary(file => file.FileName, file => (FileCoverage)file));
+            IDictionary<string, FileCoverage> lineCoverageByFile = new Dictionary<string, FileCoverage>();
```
Raphael-N, line 78: The Naming is a bit confusing here for me. What is the difference between Path and File? I can see that they are different things from the data structures behind it but I don't know why? Could you change the names or add some comments to make it easier to understand?

Raphael-N (changes requested):

UploadDaemon/Report/Testwise/Test.cs
mpdeimos: I agree, changing to path

UploadDaemon/Report/Testwise/TestwiseCoverageReport.cs
```diff
@@ -11,11 +11,17 @@ namespace UploadDaemon.Report.Testwise
...
+        [JsonProperty("partial")]
+        public bool Partial { get; }
 
-        public TestwiseCoverageReport(params Test[] tests)
+        [JsonProperty("tests")]
+        public List<Test> Tests { get; }
+
+        public TestwiseCoverageReport(params Test[] tests) : this(false, tests) {}
+
+        public TestwiseCoverageReport(bool partial, params Test[] tests)
```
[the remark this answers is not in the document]
mpdeimos: sadly no, we're also calling with one element or from above constructor. But we can change the list to array in the DTO

PR #158, opened 2021-10-04
Improve testwise coverage profiling performance by custom set implementation

Raphael-N:
Replace set where methodIds are stored with a vector for performance improvements.


Profiler/CProfilerCallback.cpp
```diff
@@ -338,8 +278,6 @@ HRESULT CProfilerCallback::AssemblyLoadFinishedImplementation(AssemblyID assembl
 		return S_OK;
 	}
 
-	EnterCriticalSection(&callbackSynchronization);
```
mpdeimos, line 341: We definitely have to synchronize assembly load. These can be loaded from different threads and we were already getting "murks" in the trace due to this ;)

Profiler/CProfilerWorker.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+void CProfilerWorker::methodIdThreadLoop() {
+	// Toggle which vector to use
+	bool vectorToggle = true;
+	while (!this->shutdown) {
+		if (vector1.empty() && vector2.empty()) {
+			std::this_thread::sleep_for(std::chrono::milliseconds(10));
+			continue;
+		}
+		if (vectorToggle) {
```
mpdeimos, line 27: in theory we could just switch pointers for primary/backing vector?

mpdeimos (changes requested):

Profiler/CProfilerCallback.cpp
```diff
@@ -303,26 +263,6 @@ DWORD CProfilerCallback::getEventMask() {
 	return dwEventMask;
 }
 
-UINT_PTR CProfilerCallback::functionMapper(FunctionID functionId, BOOL* pbHookFunction) {
-	try {
-		// Disable hooking of functions unless in TIA mode
-#ifdef TIA
-		// FIXME this causes fail: callbackInstance != NULL && callbackInstance->config.isTiaEnabled();
-		* pbHookFunction = enableFunctionHooks;
```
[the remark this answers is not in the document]
Raphael-N: The functionmapper is only active, when COR_PRF_MONITOR_ENTERLEAVE is active. Otherwise it doesn't do anything
In a previous version, COR_PRF_MONITOR_ENTERLEAVE was set for TGA profiling but that was removed quite a while ago. Now with TIA it's called but all we do is return the original FunctionID and set pbHookFunction to true which is exactly what it does without this method as far as I understood.

Profiler/CProfilerWorker.h
```diff
@@ -4,34 +4,29 @@
...
+
 	// Variables
 	std::thread* workerThread = NULL;
 	TraceLog* traceLog = NULL;
 	bool shutdown = false;
-	concurrency::concurrent_vector<FunctionID>* backing = new concurrency::concurrent_vector<FunctionID>();
-	concurrency::concurrent_vector<FunctionID>* primary = new concurrency::concurrent_vector<FunctionID>();
-	std::unordered_set<FunctionID>* calledMethodIds;
+
+#pragma warning( disable : 4316)
```
mpdeimos, line 22: can you add a justification? requires googling the number :)

Profiler/Profiler.vcxproj
```diff
@@ -169,6 +169,9 @@
       <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <ExceptionHandling>Async</ExceptionHandling>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
+      <IntrinsicFunctions>true</IntrinsicFunctions>
+      <FavorSizeOrSpeed>Speed</FavorSizeOrSpeed>
```
mpdeimos, line 173: 🏎

Profiler/utils/atomic_queue/atomic_queue.h
```diff
@@ -0,0 +1,615 @@
+/* -*- mode: c++; c-basic-offset: 4; indent-tabs-mode: nil; tab-width: 4 -*- */
+#ifndef ATOMIC_QUEUE_ATOMIC_QUEUE_H_INCLUDED
+#define ATOMIC_QUEUE_ATOMIC_QUEUE_H_INCLUDED
+
+// Copyright (c) 2019 Maxim Egorushkin. MIT License. See the full licence in file LICENSE.
```
mpdeimos, line 5: I think we should add such a LICENSE file?

Profiler/utils/functionID_set/functionId_set.h
```diff
@@ -0,0 +1,150 @@
...
+
+
+	// Predetermined values for xor operation to place in different spots
+	unsigned int xor_values[10] = { 2134038170, 2362107340, 3229546752, 1302939050, 200405764, 79516981, 2052331209, 3415124361, 940592490, 430981309 };
+
+	/// <summary>
+	/// Updates the size of the array and reinserts the element into the new array.
+	/// </summary>
+	void adjust_size() {
+		current_size *= 2;
```
mpdeimos, line 28: can this yield to overflows if done unchecked?

Profiler/utils/functionID_set/functionId_set.h
```diff
@@ -0,0 +1,150 @@
...
+		max_elements = current_size / 2;
```
mpdeimos, line 29: if you switch the statements you should be fine just assigning current_size

Profiler/utils/functionID_set/functionId_set.h
```diff
@@ -0,0 +1,150 @@
...
+		FunctionID* old_set = set;
+		set = new FunctionID[current_size]{ 0 };
+		for (unsigned int i = 0; i < current_size / 2; i++) {
```
mpdeimos, line 32: `max_elements` instead of `current_size / 2`?

Profiler/utils/functionID_set/functionId_set.h
```diff
@@ -0,0 +1,150 @@
...
+		if (set[position] == 0) {
+			return false;
+		}
+		if (set[position] == f) {
+			return true;
+		}
+
+		// Then rotate bits and xor to try and find a new position
+		for (int i = 0; i < 10; i++) {
+			position = (rotr32(f) ^ xor_values[0]) & (current_size - 1);
```
mpdeimos, line 93: shouldn't this be `xor_values[i]`?

Profiler/utils/functionID_set/functionId_set.h
```diff
@@ -0,0 +1,150 @@
...
+		unsigned int position = f & (current_size - 1);
+		if (set[position] == 0) {
+			return false;
+		}
+		if (set[position] == f) {
+			return true;
+		}
+
+		// Then rotate bits and xor to try and find a new position
+		for (int i = 0; i < 10; i++) {
```
mpdeimos, line 92: extract 10 to const? also for size of `xor_values`

Profiler/utils/functionID_set/functionId_set.h
```diff
@@ -0,0 +1,150 @@
...
+		unsigned int position = f & (current_size - 1);
+		if (set[position] == f) {
+			return;
+		}
+		if (set[position] == 0) {
+			set[position] = f;
+			return;
+		}
+
+		for (int i = 0; i < 10; i++) {
```
mpdeimos, line 130: constant

Profiler/utils/functionID_set/functionId_set.h
```diff
@@ -0,0 +1,150 @@
...
+			position = (f ^ xor_values[0]) & (current_size - 1);
```
mpdeimos, line 131: shouldn't this be `xor_values[i]`?

Profiler_Cpp_Test/Profiler_Cpp_Test.vcxproj
```diff
@@ -37,7 +37,7 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v141</PlatformToolset>
```
mpdeimos, line 40: is the downgrade here intentional?

Profiler/utils/functionID_set/functionId_set.h
```diff
@@ -0,0 +1,150 @@
...
+	/// <summary>
+	/// Inserts FunctionID f into the set.
+	/// </summary>
+	void insert(FunctionID f) {
+		num_elements++;
+		if (num_elements > max_elements) {
+			adjust_size();
+		}
+		unsigned int position = f & (current_size - 1);
+		if (set[position] == f) {
```
mpdeimos, line 122: extract the following two if-clauses into (inlined?) method?

Profiler_Test/Tia/TiaProfilerTest.cs
```diff
@@ -142,15 +142,15 @@ public void IpcStartedAfterStartup()
...
+            profilerIpc.Dispose();
             TesteeProcess testeeProcess = Start(testee, profilerUnderTest);
 
-            profilerIpc = CreateProfilerIpc(oldProfilerIpc.Config);
+            profilerIpc = CreateProfilerIpc(profilerIpc.Config);
             RunTestCase("A", testeeProcess, profilerIpc);
             Stop(testeeProcess);
 
-            Assert.That(oldProfilerIpc.ReceivedRequests, Is.Empty);
+            //Assert.That(oldProfilerIpc.ReceivedRequests, Is.Empty);
```
mpdeimos, line 153: I do not understand the test case anymore (i guess i've even written it). 

should we maybe assign oldProfilerIpc to profilerIpc after dispose / before starting the new one?

mpdeimos (changes requested):
Thanks. The usual comments, but I think we're getting there :+1:

Profiler/utils/functionID_set/functionId_set.h
mpdeimos: then maybe just throw an exception as well? I just fear that this way we end up with corrupted memory.

Raphael-N: I did a bit of testing and when it cannot allocate the memory (at least when `new` is used), it throws std::bad_alloc and stops the application. That should be enough, I guess.

PR #165, opened 2022-05-30
TS-30302 Replaced line coverage by method-accurate coverage in userguide

j-christl:
Addresses issue [TS-30302](https://jira.cqse.eu/browse/TS-30302)

- [ ] Tests written
- [ ] Documentation updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/documentation/userguide.md
+++ b/documentation/userguide.md
@@ -5,7 +5,11 @@ Test Gap analysis is a tool, which brings more transparency to the testing proce
 * the source code at the current revision
 * execution information, which is collected via the .NET Profiler
 
-The profiler writes method coverage information of a .NET application into a report file (a.k.a. *trace file*). It uses the builtin .NET profiling interface and therefore can be used with any .NET application.
+The profiler writes method-accurate coverage information of a .NET application into a report file (a.k.a. *trace file*). It uses the builtin .NET profiling interface and therefore can be used with any .NET application.
+
+Please note that the profiler cannot create line-accurate coverage, but only method-accurate coverage.
+Hence, the generated coverage information does not map to specific line numbers, but to methods.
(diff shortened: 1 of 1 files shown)


documentation/userguide.md
```diff
@@ -306,6 +306,9 @@ uploader and in Teamscale. Patterns are glob patterns: `*` matches any number of
       include: [ "*YourAssembly*" ]
       exclude: [ "*DoNotProfileThisAssembly*" ]
 
+Please note that the profiler cannot create line-accurate coverage, but only method-accurate coverage.
+The generated coverage will therefore not map to specific line numbers, but to methods.
```
saileru, line 310: I would probably move these sentences somewhere to the beginning of the file as this is essential information. This should probably be placed so that users can easily see it to prevent confusion. I would also add a reference to TGA and explain that method-accurate coverage is sufficient as TGA also operates on basis of methods instead of single lines. My suggestion would be to have this after the part "The profiler writes method coverage information of..." in the first section (lines 8ff):
```
The profiler writes method coverage information of a .NET application into a report file (a.k.a. *trace file*). It uses the builtin .NET profiling interface and therefore can be used with any .NET application.
Please note that the profiler does not create line-accurate coverage, but only method-accurate coverage.
The generated coverage will therefore not map to specific line numbers, but to methods.
This is sufficient for TGA in Teamscale which operates on the basis of methods.
```

saileru (changes requested):

documentation/userguide.md
j-christl: Great idea, thanks! Moved this and also added a sentence regarding the relation to TGA.

saileru (approved):

PR #171, opened 2023-03-28
TS-33891 ensure target directories are created

mpdeimos:
Addresses issue [TS-33891](https://jira.cqse.eu/browse/TS-33891)

### DIFF
--- a/Profiler/utils/WindowsUtils.cpp
+++ b/Profiler/utils/WindowsUtils.cpp
@@ -89,8 +89,8 @@ bool WindowsUtils::ensureDirectoryExists(std::string directory) {
 		return true;
 	}
 
-	std::size_t indexOfSeparator = directory.find_last_of("\\/");
-	if (directory.find_last_of("\\/") != std::wstring::npos) {
+	std::size_t indexOfSeparator = directory.find_last_of(separators);
+	if (indexOfSeparator != std::wstring::npos) {
 		if (!ensureDirectoryExists(directory.substr(0, indexOfSeparator))) {
 			return false;
 		}
--- a/Profiler/utils/WindowsUtils.cpp
+++ b/Profiler/utils/WindowsUtils.cpp
@@ -1,4 +1,5 @@
 #include "WindowsUtils.h"
+#include "Debug.h"
 #include <Windows.h>
 #include <vector>
 #include <algorithm>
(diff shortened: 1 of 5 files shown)


UploadDaemon/Upload/FileSystemUpload.cs
```diff
@@ -48,6 +49,15 @@ public Task<bool> UploadAsync(string filePath, string version)
...
         }
 
+        private void EnsureTargetDirectoryExists(string targetDirectory)
+        {
+            if (!Directory.Exists(targetDirectory))
+            {
+                logger.Debug("Creating target directory: {targetDirectory}", targetDirectory);
+                Directory.CreateDirectory(targetDirectory);
+            }
+        }
```
andipunz, line 59: Do we also need some handling if the directory path is already a file path as we have in the profiler itself?

Profiler/utils/WindowsUtils.cpp
```diff
@@ -71,4 +72,30 @@ bool WindowsUtils::isFile(std::string path)
...
+		return false;
+	}
+
+
+	if (dwAttrib != INVALID_FILE_ATTRIBUTES && (dwAttrib & FILE_ATTRIBUTE_DIRECTORY || dwAttrib & FILE_ATTRIBUTE_REPARSE_POINT)) {
+		// directory or junction exists
+		return true;
+	}
+
+	std::size_t indexOfSeparator = directory.find_last_of("\\/");
```
snowbagoly, line 92: Could you use the `separators` here? (otherwise it looks unused)

UploadDaemon/Upload/FileSystemUpload.cs
mpdeimos: Good point.

Profiler/utils/WindowsUtils.cpp
mpdeimos: yep

Profiler/utils/WindowsUtils.cpp
```diff
@@ -71,4 +72,30 @@ bool WindowsUtils::isFile(std::string path)
...
+	if (directory.find_last_of("\\/") != std::wstring::npos) {
```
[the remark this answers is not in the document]
mpdeimos: yep

PR #174, opened 2023-03-30
TS-33933 provide config file on startup

mpdeimos:
Addresses issue [TS-33933](https://jira.cqse.eu/browse/TS-33933)


snowbagoly (commented):
Tried out the new flags, and they seem to work nicely. 😊 I've added a few suggestions.

PR #179, opened 2023-07-04
TS-34893 Naive attempt to migrate to GitHub Actions

DreierF:
Addresses issue [TS-34893](https://jira.cqse.eu/browse/TS-34893)

- [ ] Tests written
- [x] Documentation updated
- [ ] `CHANGELOG.md` updated


.github/workflows/actions.yml
```diff
@@ -0,0 +1,84 @@
...
+
+on:
+  push:
+    branches: '**'
+    tags: 'v*'
+  pull_request:
+    branches: '**'
+
+env:
+  BUILD_VERSION: 23.6.0.${{ github.run_number }}
```
andipunz, line 11: Could we maybe use the github.ref_name?  See https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
`BUILD_VERSION: ${{ github.ref_name }}.${{ github.run_number }}`

.github/workflows/actions.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Patch C++ Version Info
+        shell: pwsh
+        run: |
+          $content = Get-Content Profiler/version.h
+          $newContent = $content -replace '1\.0\.0\.0', "${env:BUILD_VERSION}"
+          $newContent = $newContent -replace '1,0,0,0', "${env:BUILD_VERSION.Replace('.', ',').Split('-')[0]}"
+          $newContent | Set-Content Profiler/version.h
+
+      - name: Build
+        run: msbuild Cqse.Teamscale.Profiler.Dotnet.sln /p:Platform=Win32 -property:Configuration=Release
```
andipunz, line 38: I think we patched only the version for the c++ assemblies, but not for the c# assembly infos in:

- UploadDaemon/Properties/AssemblyInfo.cs
- DumpPdb/Properties/AssemblyInfo.cs

This could be easily done for dotnet build with `dotnet build SolutionName.sln -c Release /property:Version=1.2.3.4`
However I think we cannot use dotnet build right now correct?

.github/workflows/actions.yml
snowbagoly: Thanks for adding a comment here, I realized that my fishy fix in https://github.com/cqse/teamscale-profiler-dotnet/pull/179/commits/44ba4a34a85674d3522e3d5f848938e7d73de75c on the `resource.rc` is incorrect 🙈 
The problem is that we should ensure that the BUILD_VERSION is in the format of `{2-byte-number}.{2-byte-number}.{2-byte-number}.{2-byte-number}` which might already not be the case (as it was breaking the builds after the "Patch C++ Version Info" step). I will take a look.

.github/workflows/actions.yml
andipunz: Update, I think we can use this Github Action for that: https://github.com/secondbounce/assemblyinfo-update

snowbagoly: AFAIU `msbuild` might also be able to pass that property, I will try it out locally. 👍

snowbagoly: `msbuild` would need some custom scripting to work. I also tried to do the build with `dotnet build` locally but it ran into some weird issues. :grin:

Thanks for looking up the GitHub Action, it works great! 😊

.github/workflows/actions.yml
```diff
@@ -0,0 +1,109 @@
...
+        run: |
+          if (-not ($env:BUILD_VERSION -match "^v((\d+).(\d+).(\d+).(\d+))$")) {
+            Write-Host "Invalid version number: $env:BUILD_VERSION"
+            exit 1
+          }
+
+          echo "BUILD_VERSION=$($matches[1])" >> $env:GITHUB_ENV
+
+      - name: Set version string to 1.0.0.0 for non-release
+        if: "!startsWith(github.ref, 'refs/tags/v')"
```
snowbagoly, line 30: FYI quotes necessary as otherwise syntax is incorrect :grin: (https://github.com/orgs/community/discussions/25017)

.github/workflows/actions.yml
snowbagoly: It turned out that the Powershell expression was a bit off and set the version to empty string, now it should be fixed. :grin:

I applied the change on calculating the `BUILD_VERSION`, and added a check so the build fails early if the version number is off. It also cuts off the `v` prefix of the tag (there is no built-in GitHub Action expression for it, see https://github.com/orgs/community/discussions/26625)

PR #189, opened 2024-08-09
TS-30100 add new sample project for debugging

jveihelmann:
Addresses issue [TS-30100](https://jira.cqse.eu/browse/TS-30100)

- [ ] Tests written
- [x] Documentation updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/Sample_Debugging_App/Properties/launchSettings.json
+++ b/Sample_Debugging_App/Properties/launchSettings.json
@@ -1,0 +1,13 @@
+{
+  "profiles": {
+    "Sample_Debugging_App": {
+      "commandName": "Project",
+      "environmentVariables": {
+        "CORECLR_ENABLE_PROFILING": "1",
+        "CORECLR_PROFILER": "{DD0A1BB6-11CE-11DD-8EE8-3F9E55D89593}",
+        "CORECLR_PROFILER_PATH_64": "C:\\Dev\\teamscale-profiler-dotnet\\Profiler\\bin\\Debug\\Profiler64.dll"
+      },
+      "nativeDebugging": true
+    }
+  }
+}
--- a/Sample_Debugging_App/Properties/launchSettings.json
+++ b/Sample_Debugging_App/Properties/launchSettings.json
@@ -5,7 +5,7 @@
       "environmentVariables": {
         "CORECLR_ENABLE_PROFILING": "1",
         "CORECLR_PROFILER": "{DD0A1BB6-11CE-11DD-8EE8-3F9E55D89593}",
-        "CORECLR_PROFILER_PATH_64": "C:\\Dev\\teamscale-profiler-dotnet\\Profiler\\bin\\Debug\\Profiler64.dll"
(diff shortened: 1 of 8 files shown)


documentation/CONTRIBUTE.md
```diff
@@ -1,27 +1,11 @@
...
+
+The `Sample Debugging App` inside this repository serves to debug the profiler.
+To get it going please follow these steps:
+	* In Visual Studio, select the `Debug` solution config and `x64` as solution platform.
+	* Build the profiler.
+	* In the Sample Debugging App properties go to the `Debug` section and select `Open debug launch settings UI`.
+	* Change `CORECLR_PROFILER_PATH_64` to the path of your `Profiler64.dll`. This path needs to be absolute and should look similar to `C:\\Dev\\teamscale-profiler-dotnet\\Profiler\\bin\\Debug\\Profiler64.dll`.
+	* Start the Sample Debugging App via Visual Studio. 
+	
+Your breakpoints in the profiler should now be working correctly and help you with debugging.
```
snowbagoly, line 11: Could you also add a point on the `Profiler.yml`? 😁 I realize that by creating a fresh git clone it does not exist, but I ran into the issue that I already had a Profiler.yml file from a previous run right next to the Profiler64.dll. This excluded the Sample Debugging App from profiling, causing that nothing happened when I was trying to debug. 😅

Could be something like: "Make sure that if a `Profiler.yml` configuration file exists in the DLLs folder, it should enable profiling for `Sample_Debugging_App.exe`"

Sample_Debugging_App/Properties/launchSettings.json
```diff
@@ -0,0 +1,13 @@
+{
+  "profiles": {
+    "Sample_Debugging_App": {
+      "commandName": "Project",
+      "environmentVariables": {
+        "CORECLR_ENABLE_PROFILING": "1",
+        "CORECLR_PROFILER": "{DD0A1BB6-11CE-11DD-8EE8-3F9E55D89593}",
+        "CORECLR_PROFILER_PATH_64": "C:\\Dev\\teamscale-profiler-dotnet\\Profiler\\bin\\Debug\\Profiler64.dll"
```
snowbagoly, line 8: Could we set the path by default to `..\..\..\..\Profiler\bin\Debug\Profiler64.dll` or too unstable? :grin: (For me the executable was under `Sample_Debugging_App\bin\Debug\net6.0`, so had to go up 4 folders :grin:)

snowbagoly (commented):
This is pretty amazing, thanks for implementing it! 🎉 I added a few questions 🙂

Sample_Debugging_App/Properties/launchSettings.json
jveihelmann: Ah that is great! Our documentation demands an "absolute path" at this point, so I wasn't sure if a relative path could work. I tried it and it works fine! :)

PR #206, opened 2026-07-31
TS-47210 Implement endpoint for coverage dump

Raphael-N:
Addresses issue [TS-XXXXX](https://jira.cqse.eu/browse/TS-XXXXX)

- [ ] Tests written
- [ ] Documentation updated
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/Commander.Server/ProfilerController.cs
+++ b/Commander.Server/ProfilerController.cs
@@ -1,0 +1,32 @@
+using Cqse.Teamscale.Profiler.Commons.Ipc;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Cqse.Teamscale.Profiler.Commander.Server
+{
+    [Route("profiler")]
+    [ApiController]
+    public class ProfilerController : ControllerBase
+    {
+        private readonly ProfilerIpc profilerIpc;
+        private readonly ILogger logger;
+
+        public ProfilerController(ProfilerIpc profilerIpc, ILogger<ProfilerController> logger)
+        {
+            this.profilerIpc = profilerIpc;
+            this.logger = logger;
+        }
+
+        /// <summary>
+        /// Asks all connected profilers to write the coverage they collected so far to disk. This only
+        /// returns once they have done so, i.e. the profiled applications may be killed afterwards
+        /// without losing their coverage.
+        /// </summary>
+        [HttpPost("dump")]
+        public IActionResult DumpCoverage()
+        {
+            logger.LogInformation("Dumping the coverage of all connected profilers");
+            profilerIpc.DumpCoverage();
+            return NoContent();
+        }
+    }
+}
--- a/Commander.Server/ProfilerController.cs
+++ b/Commander.Server/ProfilerController.cs
@@ -17,9 +17,8 @@ namespace Cqse.Teamscale.Profiler.Commander.Server
         }
 
         /// <summary>
-        /// Asks all connected profilers to write the coverage they collected so far to disk. This only
-        /// returns once they have done so, i.e. the profiled applications may be killed afterwards
-        /// without losing their coverage.
+        /// Asks all connected profilers to write the coverage they collected so far to disk. 
+        /// This is a blocking call and only returns once they are done.
         /// </summary>
         [HttpPost("dump")]
         public IActionResult DumpCoverage()
--- a/Commons/Ipc/ProfilerIpc.cs
+++ b/Commons/Ipc/ProfilerIpc.cs
@@ -8,12 +8,6 @@ namespace Cqse.Teamscale.Profiler.Commons.Ipc
     {
         private static readonly Logger logger = LogManager.GetCurrentClassLogger();
 
-        /// <summary>
-        /// The profiler only acknowledges a dump once the coverage has been handed over to the operating
-        /// system, so we allow more time for it than for a test event.
-        /// </summary>
-        private static readonly TimeSpan DumpTimeout = TimeSpan.FromSeconds(30);
-
         private readonly ZmqIpcServer ipcServer;
 
         /// <summary>
@@ -89,7 +83,7 @@ namespace Cqse.Teamscale.Profiler.Commons.Ipc
         public void DumpCoverage()
         {
             logger.Info("Requesting a coverage dump from all connected profilers");
-            ipcServer.Broadcast("dump", DumpTimeout);
+            ipcServer.Broadcast("dump");
         }
 
         public void Dispose()
--- a/Commons/Ipc/ProfilerIpc.cs
+++ b/Commons/Ipc/ProfilerIpc.cs
@@ -8,6 +8,12 @@ namespace Cqse.Teamscale.Profiler.Commons.Ipc
     {
         private static readonly Logger logger = LogManager.GetCurrentClassLogger();
 
+        /// <summary>
+        /// The profiler only acknowledges a dump once the coverage has been handed over to the operating
+        /// system, so we allow more time for it than for a test event.
+        /// </summary>
+        private static readonly TimeSpan DumpTimeout = TimeSpan.FromSeconds(30);
+
         private readonly ZmqIpcServer ipcServer;
 
         /// <summary>
@@ -73,6 +79,17 @@ namespace Cqse.Teamscale.Profiler.Commons.Ipc
             logger.Info("Broadcasting end of test {testName} with result {result}", TestName, result);
             this.TestName = string.Empty;
             ipcServer.SendTestEvent($"end:{Enum.GetName(typeof(TestExecutionResult), result).ToUpper()}:{durationMs}");
+        }
+
+        /// <summary>
+        /// Asks all connected profilers to write the coverage they collected so far to disk.
+        /// Returns once the profilers have done so, which makes it safe to kill the profiled
+        /// applications afterwards without losing their coverage.
+        /// </summary>
+        public void DumpCoverage()
+        {
+            logger.Info("Requesting a coverage dump from all connected profilers");
+            ipcServer.Broadcast("dump", DumpTimeout);
         }
 
         public void Dispose()
--- a/Commons/Ipc/ZmqIpcServer.cs
+++ b/Commons/Ipc/ZmqIpcServer.cs
@@ -157,7 +157,6 @@ namespace Cqse.Teamscale.Profiler.Commons.Ipc
         /// <summary>
         /// Drops all clients whose profiled process has ended. Profiled applications are often killed
         /// instead of being shut down gracefully, in which case they never tell us that they are gone.
-        /// Without this, every broadcast would have to wait for the response timeout of each of them.
         /// </summary>
         private void RemoveClientsOfDeadProcesses()
         {
@@ -180,14 +179,12 @@ namespace Cqse.Teamscale.Profiler.Commons.Ipc
         {
             try
             {
-                using (Process process = Process.GetProcessById(pid))
-                {
-                    return !process.HasExited;
-                }
+                using Process process = Process.GetProcessById(pid);
+                return !process.HasExited;
             }
             catch (ArgumentException)
             {
-                // thrown if no process with that id is running
+                // if no process with that id is running
                 return false;
             }
         }
--- a/Commons/Ipc/ZmqIpcServer.cs
+++ b/Commons/Ipc/ZmqIpcServer.cs
@@ -17,9 +17,23 @@ namespace Cqse.Teamscale.Profiler.Commons.Ipc
         private const string REGISTER_CLIENT = "register";
 
         /// <summary>
-        /// How long to wait for a profiler to acknowledge a broadcast message.
+        /// Message a profiler is asked to answer before it is sent an actual request
         /// </summary>
-        public static readonly TimeSpan DefaultResponseTimeout = TimeSpan.FromSeconds(3.0);
+        private const string ALIVE_CHECK = "alive";
+
+        /// <summary>
+        /// How long to wait for a profiler to answer the alive check. Answering it requires no work
+        /// from the profiler, so this stays short no matter how expensive the request that follows is.
+        /// </summary>
+        public static readonly TimeSpan AliveCheckTimeout = TimeSpan.FromSeconds(3.0);
+
+        /// <summary>
+        /// How long to wait for a profiler to acknowledge a broadcast message. This is only ever waited
+        /// for by a profiler that just answered the alive check, i.e. one that we know is there and
+        /// responsive, so it can be generous enough for the most expensive request we send. A profiler
+        /// that is gone is already sorted out by the alive check.
+        /// </summary>
+        public static readonly TimeSpan ResponseTimeout = TimeSpan.FromMinutes(1.0);
 
         private NetMQPoller? poller;
         private ResponseSocket? responseSocket;
@@ -99,17 +113,16 @@ namespace Cqse.Teamscale.Profiler.Commons.Ipc
         /// </summary>
         public void SendTestEvent(string testEvent)
         {
-            Broadcast(testEvent, DefaultResponseTimeout);
+            Broadcast(testEvent);
         }
 
         /// <summary>
         /// Sends the given message to all connected profiler instances and waits for their acknowledgement.
-        /// Profilers that don't answer within the given timeout are considered gone and are removed.
+        /// Profilers that don't answer are considered gone and are removed.
+        /// Each profiler is first asked to answer an alive check.
         /// </summary>
-        public void Broadcast(string message, TimeSpan responseTimeout)
+        public void Broadcast(string message)
         {
-            RemoveClientsOfDeadProcesses();
-
             KeyValuePair<int, ProfilerClient>[] clients;
             lock (pidToClient)
             {
@@ -125,19 +138,29 @@ namespace Cqse.Teamscale.Profiler.Commons.Ipc
             {
                 System.Threading.Tasks.Parallel.ForEach(clients, entry =>
                 {
-                    entry.Value.Socket.SendFrame(Encoding.UTF8.GetBytes(message));
-                    if (entry.Value.Socket.TryReceiveFrameString(responseTimeout, out string? response))
-                    {
-                        logger.Info($"Got Response from {entry.Value.ClientAddress}: {response}");
-                    }
-                    else
+                    void Remove()
                     {
                         lock (clientsToRemove)
                         {
                             clientsToRemove.Add(entry.Key);
                         }
-                        logger.Error($"Got no response from Profiler with PID {entry.Key} with address {entry.Value.ClientAddress}, removing from clients");
                     }
+
+                    if (Request(entry.Value, ALIVE_CHECK, AliveCheckTimeout) == null)
+                    {
+                        Remove();
+                        logger.Error($"Profiler with PID {entry.Key} with address {entry.Value.ClientAddress} did not answer the alive check within {AliveCheckTimeout}, removing from clients without sending '{message}'");
+                        return;
+                    }
+
+                    string? response = Request(entry.Value, message, ResponseTimeout);
+                    if (response == null)
+                    {
+                        Remove();
+                        logger.Error($"Profiler with PID {entry.Key} with address {entry.Value.ClientAddress} answered the alive check but did not acknowledge '{message}' within {ResponseTimeout}, removing from clients");
+                        return;
+                    }
+                    logger.Info($"Got Response from {entry.Value.ClientAddress}: {response}");
                 });
             }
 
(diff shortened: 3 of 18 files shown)


Profiler_Test/Tia/TiaProfilerTest.cs
```diff
@@ -156,6 +155,18 @@ public void IpcStartedAfterStartup()
...
         }
 
+        /// <summary>
+        /// Waits until the profiler has registered at the given IPC server. A profiler that starts
+        /// without a commander looks for one in the background, so it only notices an IPC server that
+        /// starts later on its next attempt.
+        /// </summary>
+        private static void WaitForProfilerToRegister(RecordingProfilerIpc profilerIpc)
+        {
+            // the profiler asks for the currently running test right after it registered
```
snowbagoly: We don't need this comment; you can move it to the doc of the method if you would like to keep it

Profiler_Test/Tia/TiaProfilerTest.cs
```diff
@@ -156,6 +155,18 @@ public void IpcStartedAfterStartup()
             Assert.That(testResult.TestCases[1].TraceLines, Has.Some.Matches("^(Inlines|Jitted|Called)"));
         }
 
+        /// <summary>
+        /// Waits until the profiler has registered at the given IPC server. A profiler that starts
```
snowbagoly, line 159: Maybe you can shorten this: currently it explains how it notices the IPC server, but  for me does not really connect the things together. You could describe that "we might need several retries if the IPC server starts later than the profiler" or something like that.

Profiler/CProfilerCallback.cpp
```diff
@@ -138,19 +138,33 @@ namespace Profiler {
 			createDaemon().launch(traceLog);
 		}
 
+		// The profiler always connects to the commander, even in TGA mode, where it is used to request
+		// that the collected coverage is written to disk.
```
snowbagoly: We probably don't need this comment

Profiler/CProfilerCallback.cpp
```diff
@@ -138,19 +138,33 @@ namespace Profiler {
...
+		}
+		else {
+			traceLog.info("IPC REQ Socket: " + config.getTiaRequestSocket());
+		}
+		std::function<void(std::string)> testStartCallback = [this](std::string testName) {
+			this->onTestStart(testName);
+			};
+		std::function<void(std::string, std::string)> testEndCallback = [this](std::string result, std::string duration) {
+			this->onTestEnd(result, duration);
+			};
```
snowbagoly: Should we register these only for the TIA case? Would it fail if we send a test start request otherwise?

Profiler/CProfilerCallback.cpp
```diff
@@ -138,19 +138,33 @@ namespace Profiler {
...
+		std::function<void(std::string)> infoCallback = [this](std::string message) {
+			this->traceLog.info(message);
+			};
+		std::function<void(std::string)> errorCallback = [this](std::string message) {
+			this->traceLog.error(message);
+			};
+		this->ipc = std::make_unique<Ipc>(&this->config, testStartCallback, testEndCallback, dumpCallback, infoCallback, errorCallback);
 
+		if (config.isTiaEnabled()) {
 			setCriticalSection(&methodSetSynchronization);
```
snowbagoly: Can we merge this with the other `config.isTiaEnabled` section?

documentation/userguide.md
```diff
@@ -222,6 +222,39 @@ Configuration options from environment variables always override configuration o
 Please note that you **cannot** register the profiler itself via the config file (`COR_PROFILER`, `COR_ENABLE_PROFILING`).
 
 
+# Dumping Coverage on Request
```
snowbagoly, line 225: Please move these into the Teamscale docs on 2026.5.x or 2026.6.x (depending on when you expect that we will merge this 😂 ) I will still have to do the same with my changes 😬

Profiler/CProfilerCallback.cpp
```diff
@@ -533,6 +547,22 @@ namespace Profiler {
 			<< version[3];
 	}
 
+	void CProfilerCallback::onDumpRequest()
+	{
+		if (!config.isProfilingEnabled()) {
+			return;
+		}
+		EnterCriticalSection(&callbackSynchronization);
```
snowbagoly: As discussed, we don't know if this `callbackSynchronization` should be entered. If yes, we should probably also add it to the test start and test end methods.

Profiler/utils/Ipc.h
```diff
@@ -23,16 +27,58 @@ namespace Profiler {
...
+
+		/** Whether we ever successfully registered with the commander. */
+		bool isRegistered = false;
+
+		/** Used to abort the wait between two registration attempts as soon as the profiler shuts down. */
+		std::mutex shutdownMutex;
+		std::condition_variable shutdownCondition;
+
+		/** Used to jitter the interval between two registration attempts. */
+		std::mt19937 randomEngine;
```
snowbagoly: You can extend the comment that this was introduced for the case when many profilers are running at once - or revert this change if we don't expect this to happen.

Commander.Server/ProfilerController.cs
```diff
@@ -0,0 +1,32 @@
...
+        public ProfilerController(ProfilerIpc profilerIpc, ILogger<ProfilerController> logger)
+        {
+            this.profilerIpc = profilerIpc;
+            this.logger = logger;
+        }
+
+        /// <summary>
+        /// Asks all connected profilers to write the coverage they collected so far to disk. This only
+        /// returns once they have done so, i.e. the profiled applications may be killed afterwards
+        /// without losing their coverage.
```
snowbagoly: I think "the profiled applications may be killed afterwards without losing their coverage" is a bit of a strong claim: as long as they are running, they might have additional coverage, we can't guarantee that everything is written.

I would keep it to
```
       /// Asks all connected profilers to write the coverage they collected so far to disk. This is a blocking call and only returns once they are done.
```

Profiler/utils/Ipc.cpp
```diff
@@ -126,6 +215,16 @@ namespace Profiler {
 		return true;
 	}
 
+	void Ipc::waitForRetry(std::chrono::milliseconds duration) {
+		std::unique_lock<std::mutex> lock(this->shutdownMutex);
+		this->shutdownCondition.wait_for(lock, duration, [this] { return this->shutdown.load(); });
```
snowbagoly, line 208: Maybe iterate once with Claude if this can accidentally starve the destructor or Claude is confident that this cannot happen. 🤷

Profiler/utils/Ipc.h
```diff
@@ -23,16 +27,58 @@ namespace Profiler {
...
 		std::function<void(std::string, std::string)> testEndCallback;
+		std::function<void()> dumpCallback;
+		std::function<void(std::string)> infoCallback;
 		std::function<void(std::string)> errorCallback;
 		std::atomic<bool> shutdown = false;
+
+		/** Whether we ever successfully registered with the commander. */
+		bool isRegistered = false;
+
+		/** Used to abort the wait between two registration attempts as soon as the profiler shuts down. */
```
snowbagoly: Can we rephrase it to something like "Blocking attempts to connect to the Commander while shutting down" if it's for that specifically?

Profiler/utils/Ipc.cpp
```diff
@@ -126,6 +215,16 @@ namespace Profiler {
 		return true;
 	}
 
+	void Ipc::waitForRetry(std::chrono::milliseconds duration) {
```
snowbagoly: Can we rename this to `wait` or `waitBeforeRetry`, so I don't assume that we are doing something with the retry here?

Profiler/utils/Ipc.cpp
```diff
@@ -126,6 +215,16 @@ namespace Profiler {
...
+		std::unique_lock<std::mutex> lock(this->shutdownMutex);
+		this->shutdownCondition.wait_for(lock, duration, [this] { return this->shutdown.load(); });
+	}
+
+	std::chrono::milliseconds Ipc::nextRetryInterval() {
```
snowbagoly: Can we inline this one?

Profiler/utils/Ipc.h
```diff
@@ -23,16 +27,58 @@ namespace Profiler {
...
+		 */
+		std::chrono::milliseconds nextRetryInterval();
+
+		/**
+		 * Sends the given message to the commander and returns its answer, or an empty string if the
+		 * commander could not be reached or did not answer in time. If commanderWasReachable is given,
+		 * it is set to whether we had a connection to the commander at all, which tells a missing
+		 * commander apart from one that is there but does not answer.
+		 */
+		std::string request(const std::string& message, bool* commanderWasReachable = nullptr);
```
snowbagoly: Can we have a compound return value `string` and `bool` instead of the pointer?

Commons/Ipc/ProfilerIpc.cs
```diff
@@ -8,6 +8,12 @@ public class ProfilerIpc : IDisposable
     {
         private static readonly Logger logger = LogManager.GetCurrentClassLogger();
 
+        /// <summary>
+        /// The profiler only acknowledges a dump once the coverage has been handed over to the operating
+        /// system, so we allow more time for it than for a test event.
+        /// </summary>
+        private static readonly TimeSpan DumpTimeout = TimeSpan.FromSeconds(30);
```
snowbagoly: Should we make this configurable?

Commons/Ipc/ZmqIpcServer.cs
```diff
@@ -86,22 +99,48 @@ private void RegisterClient(string message)
...
         public void SendTestEvent(string testEvent)
         {
+            Broadcast(testEvent, DefaultResponseTimeout);
+        }
+
+        /// <summary>
+        /// Sends the given message to all connected profiler instances and waits for their acknowledgement.
+        /// Profilers that don't answer within the given timeout are considered gone and are removed.
+        /// </summary>
+        public void Broadcast(string message, TimeSpan responseTimeout)
```
snowbagoly: Can we have a "short" reply from the profilers before they start dumping? E.g. "You still alive?" "Sure" "Pls dump" "On it boss"
Just in case they are actually still alive, just taking long.

Commons/Ipc/ZmqIpcServer.cs
```diff
@@ -115,6 +154,41 @@ public void SendTestEvent(string testEvent)
...
+        /// Drops all clients whose profiled process has ended. Profiled applications are often killed
+        /// instead of being shut down gracefully, in which case they never tell us that they are gone.
+        /// </summary>
+        private void RemoveClientsOfDeadProcesses()
+        {
+            lock (pidToClient)
+            {
+                foreach (int pid in pidToClient.Keys.ToList())
+                {
+                    if (IsProcessRunning(pid))
```
snowbagoly: This only works if they are running on the same machine.

Profiler/utils/Ipc.cpp
```diff
@@ -9,44 +9,100 @@ namespace Profiler {
...
+		// wake the handler thread if it is waiting for the next registration attempt so we don't
+		// delay the shutdown of the profiled process
+		this->shutdownCondition.notify_all();
 		if (this->handlerThread->joinable()) {
 			this->handlerThread->join();
 		}
-		this->request("profiler_disconnected");
+		if (this->isRegistered) {
+			// if we never reached a commander there is nobody to tell, and trying would only wait
+			// for the connect timeout again
```
snowbagoly: That's only for the reviewer -> comment can be deleted

Profiler/utils/Ipc.cpp
```diff
@@ -90,11 +151,29 @@ namespace Profiler {
...
 	{
+		if (commanderWasReachable != nullptr) {
+			*commanderWasReachable = false;
+		}
 		if (!initRequestSocket()) {
 			return "";
 		}
+
+		// Because of ZMQ_IMMEDIATE the socket only becomes writable once the connection to the commander
+		// has actually been established, so this tells us whether there is a commander at all. Without
```
snowbagoly: Define "this": "so ...inserttexthere... tells us"

Profiler/utils/Ipc.cpp
```diff
@@ -90,11 +151,29 @@ namespace Profiler {
...
+		// it we would send into the void and only notice when the receive below times out. We have to
+		// wait a moment because ZeroMQ establishes the connection asynchronously.
+		zmq_pollitem_t pollItem{ this->zmqRequestSocket, 0, ZMQ_POLLOUT, 0 };
+		if (zmq_poll(&pollItem, 1, CONNECT_TIMEOUT_MS) <= 0) {
+			// keep the socket, ZeroMQ keeps trying to establish the connection in the background.
+			// A message we never sent also leaves the REQ socket ready to send again.
```
snowbagoly: We didn't try to send a message yet, did we? Then the second part of the comment might not be relevant.

Profiler/utils/Ipc.cpp
```diff
@@ -116,6 +195,16 @@ namespace Profiler {
...
 			zmq_setsockopt(this->zmqRequestSocket, ZMQ_RCVTIMEO, &IPC_TIMEOUT_MS, sizeof(IPC_TIMEOUT_MS));
 			zmq_setsockopt(this->zmqRequestSocket, ZMQ_LINGER, &IPC_TIMEOUT_MS, sizeof(IPC_TIMEOUT_MS));
+			// Only queue messages for connections that are established. This lets a send tell us whether
+			// the commander is reachable, instead of ZeroMQ silently queueing for a peer that may never
+			// appear. Must be set before connecting.
+			constexpr int immediate = 1;
+			zmq_setsockopt(this->zmqRequestSocket, ZMQ_IMMEDIATE, &immediate, sizeof(immediate));
+			// Let ZeroMQ retry the connection at the same jittered rate at which we look for the
+			// commander. The default of 100ms would have its I/O thread reconnecting far more often
+			// than we ask, which is wasted work for the many processes profiled without a commander.
```
snowbagoly: This comment might not be needed, except if you would like to explain the part that the default of 100 ms is too often.

Profiler/utils/Ipc.cpp
```diff
@@ -9,44 +9,100 @@ namespace Profiler {
...
+				if (commanderWasReachable) {
+					// somebody is there but the handshake didn't work, which is a real problem
+					logError("Registration at the commander failed, trying again.");
+				}
+				else {
+					// there simply is no commander, which is the normal case when profiling without one
+					infoCallback("No commander is listening at " + this->config->getTiaRequestSocket()
+						+ ". Looking for one in the background.");
+				}
+				connectionErrorLogged = true;
```
snowbagoly: `connectionErrorLogged` -> `connectionHintLogged`

Profiler/utils/Ipc.cpp
```diff
@@ -9,44 +9,100 @@ namespace Profiler {
...
 			}
+			waitForRetry(nextRetryInterval());
+		}
+		if (!address.empty() && connectionErrorLogged) {
```
snowbagoly: Maybe `!address.empty()` is enough in the condition

documentation/userguide.md
```diff
@@ -222,6 +222,39 @@ Configuration options from environment variables always override configuration o
...
+```
+curl -X POST http://<commander-host>:<commander-port>/profiler/dump
+```
+
+Every connected profiler then appends everything it has recorded so far to its trace file and flushes it to disk. The
+trace file stays open and profiling continues into the same file, so the applications can be killed at any point
+afterwards without losing the dumped coverage. The request only returns once the profilers have written their
+coverage, i.e. it is safe to kill the applications as soon as it comes back.
+
+This requires no configuration: the profiler always connects to the commander, in TGA mode as well as in TIA mode. If
```
snowbagoly, line 243: Should we mention that you have to start up the commander additionally?

documentation/userguide.md
```diff
@@ -222,6 +222,39 @@ Configuration options from environment variables always override configuration o
...
+no commander is running, the profiler notes that once in the trace file and keeps looking for one in the background,
+at negligible cost. It connects as soon as a commander shows up, even if that is long after the application started,
```
snowbagoly, line 245: Maybe we can skip the "at negligible cost"

documentation/userguide.md
```diff
@@ -222,6 +222,39 @@ Configuration options from environment variables always override configuration o
...
+and records this in the trace file as well:
+
+```
+Info=No commander is listening at tcp://127.0.0.1:7145. Looking for one in the background.
+Info=Connected to the commander at tcp://127.0.0.1:7145
+```
+
+The alternative to this is the `COR_PROFILER_EAGERNESS` option, which writes the trace continuously after a given
+number of recorded methods. Requesting a dump is preferable when you know the points in time at which the coverage
+must be safe, as it does not write during normal operation.
```
snowbagoly, line 255: You mentioned that you would like to rework this last sentence.

documentation/userguide.md
```diff
@@ -293,6 +326,7 @@ It provides the following endpoints:
 | test/start/{testName} | POST   | Starts the test with the given name                                                                                                                                                        |
 | test/stop/{result}    | POST   | Stops the currently active test with the given result. Possible values are Passed, Ignored, Skipped, Failure, Error                                                                        |
 | test/end/{testName}   | POST   | Stops the test with the given name if it is currently active. This is a legacy endpoint and the test/stop endpoint should be preferred. Expects a test result in the body with key Result. |
+| profiler/dump         | POST   | Writes the coverage collected so far to disk in all connected profilers. Unlike the endpoints above this is not tied to a test and also works in TGA mode. See "Dumping Coverage on Request". |
```
snowbagoly, line 329: You will probably also want to add a link on "Dumping Coverage on Request" when moving it over to Teamscale.

snowbagoly (commented):
Thanks for implementing this! Here are some comments as discussed.

Profiler_Test/Tia/TiaProfilerTest.cs
Raphael-N: Removed.

Profiler_Test/Tia/TiaProfilerTest.cs
Raphael-N: Adjusted a bit.

Profiler/CProfilerCallback.cpp
Raphael-N: Removed

Profiler/CProfilerCallback.cpp
Raphael-N: I introduced a struct for the callbacks and do only initialize the tia callbacks when relevant. Also a check if they are present. An explicit nullable would be nice but I (and claude) found no good way to express this.

Profiler/CProfilerCallback.cpp
Raphael-N: Done.

Profiler/CProfilerCallback.cpp
Raphael-N: I added both critical section in both cases.
Claude suggested extracting the guarding into its own component so it doesn't have to be done 6 times which sounded reasonable.

Profiler/utils/Ipc.h
Raphael-N: I removed the jitter stuff. I don't think we need this for the number of profilers we expect. Can still reintroduce later if needed.

Profiler/utils/Ipc.cpp
Raphael-N: Claude says it's as suspected: The mutex can interrupt the waiting and trigger the shutdown right away instead of waiting for the sleep to end.

Profiler/utils/Ipc.h
Raphael-N: Added a struct for the return type.

Commons/Ipc/ProfilerIpc.cs
Raphael-N: The Commander Server currently does not have any configuration options. Maybe that changes with your PR? 
Would you introduce a config mechanism here? 
I would for now just set it to a higher timeout combined with the alive check.

Commons/Ipc/ZmqIpcServer.cs
```diff
@@ -155,24 +178,18 @@ public void Broadcast(string message, TimeSpan responseTimeout)
         }
 
         /// <summary>
-        /// Drops all clients whose profiled process has ended. Profiled applications are often killed
-        /// instead of being shut down gracefully, in which case they never tell us that they are gone.
+        /// Sends the given message to the given profiler and returns its answer, or null if it did not
+        /// answer within the given timeout. In the latter case the socket must not be used again: a
+        /// request socket that is still waiting for an answer cannot send the next message.
         /// </summary>
-        private void RemoveClientsOfDeadProcesses()
```
Raphael-N, line 161: The replacement looks weird but it was just removing this method due to the local issue.
I think, the alive check is good enough for now and maybe we can get away without this extra step.

Profiler/utils/Ipc.cpp
```diff
@@ -15,24 +16,18 @@ namespace Profiler {
 	/** How often to look for the commander while we are not registered yet. */
 	constexpr auto REGISTRATION_RETRY_INTERVAL = 1000ms;
 
-	/** Upper bound of the random amount that is added to the interval between two registration attempts. */
-	constexpr int REGISTRATION_RETRY_JITTER_MS = 500;
```
Raphael-N, line 19: Removed the whole jitter stuff.

